PR #18, opened 2019-07-09
update ssh-keygen documentation

wdpypere:
- changes `ssh-keygen` to `ssh-keygen -t rsa -b 4096`
 - update putty screenshots
 - fix some random typo's


source/access/generating_keys_with_openssh.rst
```diff
@@ -82,12 +82,12 @@ time you want to access the cluster or transfer your files.
 
 Keys are required in the OpenSSH format.
 
-If you have a public key ``id_rsa_2048_ssh.pub`` in the SSH2 format,
+If you have a public key ``id_rsa_4096_ssh.pub`` in the SSH2 format,
```
wpoely86: How can a user know this? How do you check the correct format?

wdpypere: a current key can be checked with a command like:
```
[wdpypere@scyther ~]$ ssh-keygen -l -f ~/.ssh/id_rsa.pub
4096 SHA256:ILzWV9AZdpkK0smmAOaNce2J0ka8W0r7VQlf3DmEg9c wdpypere@scyther2019 (RSA)
```

As to the documentation about SSHv1 keys, I don't really know who this is for. SSH v1 has been deprecated for a very long time. I can't even generate an SSHv1 key any more with ssh-keygen to test. I am presuming this part of the documentation is no longer needed, but I am not the correct person to make those decisions.

klust: There is also an issue with keys generated this way with some programs that do not yet accept keys in the current OpenSSH format (which in fact is their own format), but I suggest to deal with that in the documentation of the programs that have problems with those key files unless we would identify too many of those programs. Eclipse definitely is one of them, but I'm working on adding some information to the documentation that I wrote some years ago.

PR #25, opened 2019-08-02
Restructured pages on getting access

gjbex:
We noticed that it is currently too hard to find relevant info on the account request procedure.  Specifically, the pages on generating SSH keys are well hidden, and require three click-throughs.

* Eliminate one level of redirection.
* Emphasize procedural links.
* Link to information that is on the VSC website, eliminate duplication.

Feel free to comment.

### DIFF
--- a/source/access/getting_access.rst
+++ b/source/access/getting_access.rst
@@ -1,11 +1,12 @@
 Getting access
 ==============
 
-The VSC account
----------------
+VSC accounts
+------------
 
 In order to use the infrastructure of the VSC, you need a VSC-userid,
-also called a VSC account.
+also called a VSC account.  Check the `VSC website <eligible users_>`_
+for conditions.
 
 All VSC-accounts start with the letters \\"vsc\" followed by a
 five-digit number. The first digit gives information about your home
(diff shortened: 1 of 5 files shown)


source/access/getting_access.rst
```diff
@@ -1,57 +1,213 @@
...
+also called a VSC account.  Check the `VSC website <eligible users_>`_
+for conditions.
+
+All VSC-accounts start with the letters \\"vsc\" followed by a
+five-digit number. The first digit gives information about your home
+institution. There is no relationship with your name, nor is the
+information about the link between VSC-accounts and your name publicly
+accessible.
+
+Your VSC account gives you access to most the VSC Tier-2 infrastructure, though
```
smoors: access to most __of__ the VSC Tier-2 infrastructure

source/access/getting_access.rst
```diff
@@ -1,57 +1,213 @@
...
+for some more specialised hardware you have to ask access separately.
+The rationale is that  we want to ensure that that (usually rather expensive)
```
smoors: we want to ensure that __this specialised__ (usually rather expensive)

source/access/getting_access.rst
```diff
@@ -1,57 +1,213 @@
...
+hardware is used efficiently for the type of applications it was purchased for.
+Contact your :ref:`local VSC coordinator<get in touch_>`_ to arrange access
```
smoors: I think this should be:
```
Contact your `local VSC coordinator <get in touch_>`_ to arrange access
```

source/access/getting_access.rst
```diff
@@ -1,57 +1,213 @@
...
+
+.. warning::
+
+   It is clear from the above that it is very important to protect your
+   private key well. Therefore:
+   
+   - You should choose a strong passphrase to protect your private key.
+   - You should not share your key pair with other users.
+   - If you have accounts at multiple supercomputer centres (or on other
+     systems that use SSH), you should seriously consider using a
```
smoors: why not just: `... you should use a different key pair for each of those accounts` ?

source/access/getting_access.rst
```diff
@@ -1,57 +1,213 @@
...
+   generated as described above.
+
+Users of Brussels University Association
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If you only need access to the VUB cluster Hydra, you don't
+necessarily need a full VSC account but can use your regular
+institute account. More information can be found on `this VUB Web
+Notes
+page <http://www.ulb.ac.be/wserv2_oratio/oratio?f_type=view&f_context=fiches&language=nl&noteid=227>`_.
```
smoors: replace the whole paragraph with this:
```
All information about the access policy is available on the `VUB
HPC documentation website <https://hpc.vub.be/documentation>`_.

-  Researchers can use the :ref:`generic procedure <generic access procedure>`.
-  Master students can also use the infrastructure for their master
   thesis work. The promotor of the thesis should first send a
   motivation to hpc@vub.ac.be and then the :ref:`generic
   procedure <generic access procedure>` should be followed to request the account.
```

source/access/getting_access.rst
gjbex: It is not clear to me why that phrase was there?  What is not accessible by default?

source/access/getting_access.rst
gjbex: This is the original formulation, we can't enforce this anyway.  I  can emphasize it a bit more.

source/access/getting_access.rst
smoors: good question.. at VUB there is no hardware that you need to ask access for.

source/access/getting_access.rst
smoors: you can leave it as is if you prefer, it was just a suggestion.

PR #40, opened 2019-08-09
Add thinking CentOS 7 quick start

gjbex:


gjbex:
Note that I've renamed the file to comply with the naming convention used throughout the documentation.  File names
* are *always* lowercase,
* don't contain -, only _.

PR #48, opened 2019-08-27
Add text on quota check

gjbex:
Brief text on how to check quota for KU Leuven $VSC_HOME/$VSC_DATA/$VSC_SCRATCH


source/access/managing_disk_usage.rst
```diff
@@ -1,105 +1,138 @@
...
-                                  
-::                                
-                                  
+
+On filesystems with quota enabled, you can check the amount of disk space that
+is available for you, and the amount of disk space that is in use by
+you.  Unfortunately, there is not a single command that will give
+you that information for all file systems in the VSC.
+
+On most systems, ``quota`` will show you for each file system
```
stdweird: is it really "most"? this does not work for gpfs, and thus in gent does nothing useful (and afaik, also in brussel @wpoely86). 
for gent (and brussel), users can check their usage and limits in the AP

source/access/managing_disk_usage.rst
```diff
@@ -1,105 +1,138 @@
...
+   directories and will show you  the quota for your ``$VSC_SCRATCH``
+   directory on that system.
+
+
+.. _mmlsquota:
+
+Total disk space on GPFS file systems
+-------------------------------------
+
+On some parallel file systems, notably ``$VSC_SCRATCH`` on KU Leuven
```
stdweird: the main issue with mmlsquota and also du is that it reports used disk space, incl replicas. so this will confuse people
it's a pity that these commands are documented
also running `du` is pretty hard on the filesystem, so we would also like to avoid any reference to it
and to make things worse, du reports aggregate of file sizes, not the actual disk usage, so people will again get confused why quota usage and du do not match

stdweird: seems that du does not aggregate filesizes, it does somethign even stranger, but nonetheless, it's not the disk usage

source/access/managing_disk_usage.rst
```diff
@@ -1,105 +1,138 @@
...
+infrastructure (Tier-1 and Tier-2), ``mmlsquota`` should be used to
+display the total disk space you use and the limits you have.
+
+These file system is a parallel file system, i.e., IBM Spectrum Scale (formerly
+GPFS)::
+
+      $ mmlsquota  --block-size auto  vol_ddn2:leuven_scratch
```
wpoely86: Next to Stijn his remarks I also don't think it's a good idea to give these kinds of command to the users. At the minimum a wrapper that hides all the details should be used.

source/access/managing_disk_usage.rst
gjbex: With respect to `du`, I agree that is not accurate and puts strain on the file system.  However, note that it is in the HPC intro documentation.  Please open this discussion there as well.

With respect to `mmlsquota`, we need some way for a user to check quota on GPFS systems, suggestions for alternatives are welcome.

source/access/managing_disk_usage.rst
gjbex: It seems to work for $VSC_HOME and $VSC_DATA on all sites, so that would be "most".

source/access/managing_disk_usage.rst
smoors: @gjbex as ward wrote below, at VUB we use a wrapper (called myquota) that does essentially this for vsc accounts:
```
function vscquota {
awk -v repli=$1 -v fileset=$2 -F: ' /USR/ { if ($23 == fileset) { current=$11; soft=$12 ; hard=$13 ; grace=$15 } }
END {
  printf "	Current data usage: %.2f GB\n", (current/repli/1024/1024)
  printf "	Maximum data usage: %.2f GB\n", (soft/repli/1024/1024)
  if (grace != "none")
    {
    print "	After", grace, "you will not be able to write more data"
    printf "	until you decrease your data usage below %.2f GB\n", (soft/repli/1024/1024)
    }
} '
}
echo "Quota for \$VSC_HOME"
/usr/lpp/mmfs/bin/mmlsquota -Y theiahome:${USER:0:6} | vscquota 2 ${USER:0:6}
echo
echo "Quota for \$VSC_DATA"
/usr/lpp/mmfs/bin/mmlsquota -Y theiadata:${USER:0:6} | vscquota 2 ${USER:0:6}
echo
echo "Quota for \$VSC_SCRATCH"
/usr/lpp/mmfs/bin/mmlsquota -Y theiascratch:${USER:0:6} | vscquota 1 ${USER:0:6}
```

source/access/managing_disk_usage.rst
stdweird: it works for kuleuven users everywhere, sure, because it's nfs. but it doesn't work for e.g. gent users 
anyway, my main point is that i don't mind that this is documented, but this is rather kuleuven specific, and that is not very clear from some parts of the text

source/access/managing_disk_usage.rst
gjbex: @smoors Nice, but it doesn't exist in Gent/Antwerp.  I've no issues adapting it for Leuven ($VSC_HOME/$VSC_DATA is NFS, not GPFS), but it doesn't make sense to document it unless it is part of the CUE.

@wpoely86 @stdweird I'm quite happy to throw out the part on `du`, but that means the users need a standard way of checking quota, so how?

stdweird: for ugent users, we tell them to use the AP.

stdweird: and we don't have a CLI yet (but that is mainly due to mapower issues)

PR #62, opened 2019-09-10
Added line on grace period.

sbecuwe:


gjbex:
Hm, we don't have grace... (pun intended).  I don't know about the other sites.

PR #66, opened 2019-09-10
fix user support

smoors:


smoors:
@gjbex good point. we plan to move all/most of our documentation to the VSC docs once it is more or less complete. in fact we already have many topics that just link to the VSC docs.

PR #71, opened 2019-09-13
Superdome documentation

gjbex:
Try to reduce confusion in Superdome documentation.

### DIFF
--- a/source/leuven/superdome_quick_start.rst
+++ b/source/leuven/superdome_quick_start.rst
@@ -35,7 +35,7 @@ and ``place``.
 ``place``
    The place determines where the logical processors are executed, and
    is ``numanode`` for Superdome.  Each ``numanode`` has 14 cores, so if you
-   want to use, e.g., 28 cores, you would specify ``lprocs=28:numanode=2``.
+   want to use, e.g., 28 cores, you would specify ``lprocs=28:place=numanode=2``.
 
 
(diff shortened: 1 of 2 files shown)


moravveji:
I see that this branch is merged now. All seems ok, but there is this whole subsection about the login (load-balancer, choice of NX, visualization node etc) which to my idea, does not belong to her. A user who wants to find more info about Genius logins should not jump to the superdome page to find his desired info.
So, I just think it would be better to place a link to the relevant login pages, instead.

moravveji:
Furthermore, it is perhaps also worthwhile to mention the MTM compiler suit for users who would like to compile their codes on superdome.

gjbex:
The information on the login nodes was there in the past as well.  It is now consistent with the other information on login to genius.

The point is that this is in fact a quickstart, not necessarily aimed at genius users, e.g., users from other universities/externals.  Now it reads as a single page, without having to move to other pages and back.

I don't know about the compiler you mention, so please enlighten me.

moravveji:
Sure ... 
The correct name of the tool is **MPT** and not MTM, my bad.
On the Superdome quick start guide (.docx) file on sharepoint (gedeelde documenten -> cursus) you can find quite detailed info about the specialized MPT compiler for intel and foss. I just thought that this could also be useful for some users.

PR #80, opened 2019-09-24
Bugfix/leuven login nodes

gjbex:
* Reformat to emphasize the node names to use for NX configuration for KU Leuven login nodes
* Replace thinking by genius for KU Leuven Tier-2 NX nodes

### DIFF
--- a/source/access/nx_start_guide.rst
+++ b/source/access/nx_start_guide.rst
@@ -8,7 +8,7 @@ Installing NX NoMachine client
 
 -  Download the enterprise version of the client from
    the `NX Client download`_ page.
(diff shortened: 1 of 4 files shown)


mselwa:
For genius login infrastructure I find the explanation of ssh/NX a little difficult "the nodes listed below can only directly be accessed using SSH, use nx.hpc.kuleuven.be as hostname in the NX client configuraton." (typo in configuration). I would rephrase it:
"the nodes listed below can only directly be accessed in terminal using SSH. To connect with GUI follow the :ref:`NX start guide <nx_start_guide.rst>`"

PR #81, opened 2019-09-25
Tier1 description update

gjbex:
The hardware description of Tier-1 is not on par with that of the Tier-2 infrastructure.  This PR addresses this partly by adding information on
* the login infrastructure, and
* the storage infrastructure.


gjbex:
Thanks, feedback taken into account.

PR #96, opened 2019-10-04
Globus for VSC users added to VSC User Docs.

avapirev:
First draft.


source/globus/globus_platform.rst
```diff
@@ -0,0 +1,70 @@
...
+
+We welcome your feedback, comments and suggestions for improving the Globus Tutorial.
+
+
+.. _access:
+
+Access
+------
+
+.. include:: access.rst
```
gjbex: Using actual includes makes this one monolithic page.  Would it makes sense to put each section in its own page?

source/globus/globus_platform.rst
```diff
@@ -0,0 +1,70 @@
...
+
+
+.. _managing_and_transferring_files:
+
+Managing and transferring files
+-------------------------------
+
+.. include:: managing_and_transferring_files.rst
```
gjbex: If you work with includes, I'd suggest to put the section headings in the files themselves, rather than in the parent document.

source/globus/access.rst
```diff
@@ -0,0 +1,24 @@
+
+Log in with your institution:
+
+Visit `www.globus.org <https://www.globus.org>`_ and click :bgrnd1:`Login` at the top of the page. On the Globus login page, choose an organization you are already registered with, such as your school or your employer. Alternatively, one could directly navigate to `globus.org/app/login <https://www.globus.org/app/login>`_.
+
+
+.. image:: img/access-login-screen.png
```
gjbex: Perhaps it is easier to keep track of images when they are in a directory with the same name as the page they are included from?  The advantage is that when you need to move a page, it is immediately clear which images should be moved.
The disadvantage is that sharing images between pages is weird with this approach, but we didn't come across this in the documentation so far.

source/globus/managing_and_transferring_files.rst
```diff
@@ -0,0 +1,40 @@
...
+-  Click on a collection. Globus will connect to the collection and display the default directory. Navigate either by typing the destination directory into the :bgrnd1:`Path` field, or browse the available directories below.
+
+
+Transferring files
+~~~~~~~~~~~~~~~~~~
+
+.. image:: img/filetransfer-2.png
+
+
+- Click :bgrnd1:`Transfer` or :bgrnd1:`Sync to...` in the command panel on the right side of the page. A new collection panel will open, with a :bgrnd1:`Transfer or Sync to` field at the top of the panel. Alternatively, use the :bgrnd1:`Panels` toggle button in the top right corner to split the directory tree panel and select a destination collection.
```
gjbex: Isn't this a single item?

source/globus/managing_and_transferring_files.rst
```diff
@@ -0,0 +1,40 @@
...
+
+- Find and choose the second collection and connect to it as you did with the first one. Click on the left first collection and select all the files to transfer. The :bgrnd1:`Start` button at the bottom of the panel will become active. Between the two :bgrnd1:`Start` buttons at the bottom of the page, the :bgrnd1:`Transfer and Sync Options` tab provides access to several options. By default, Globus verifies file integrity after transfer using checksums. Change the transfer settings if you would like. You may also enter a label for the transfer.
+
+- Click the :bgrnd1:`Start` button to transfer the selected files to the collection in the right panel. Globus will display a green notification panel - confirming that the transfer request was submitted - and add a badge to the :bgrnd1:`Activity` item in the command menu on the left of the page. You can navigate away from the File Manager, close the browser window, and even logout. Globus will optimize the transfer for performance, monitor the transfer for completion and correctness, and recover from network errors and collection downtime.
```
gjbex: If I follow along, choosing VSC KU Leuven as endpoint to copy from and KU Leuven as endpoint to copy to (that's something your average user might do to familiarize himself with this), it doesn't seem to work (e.g., to copy from data to scratch.

source/globus/local_endpoints.rst
```diff
@@ -0,0 +1,11 @@
+Users can create and manage endpoints on their personal computers via `Globus Connect Personal <https://www.globus.org/globus-connect-personal>`_. Globus Connect Personal is available for all major operating systems. Download the software and follow the installation instructions. During the setup process you will be asked to set the local endpoint. Each endpoint has a unique ID. Those ones created or managed by you are associated with your Globus account.
```
gjbex: Since this link would be used in multiple files, I'd suggest to put in a `links.rst` file that can be included at the bottom each each file you want to use the link in.

source/globus/local_endpoints.rst
```diff
@@ -0,0 +1,11 @@
...
+
+To access and manage your local endpoints:
+
+- Start the **Globus Connect Personal** applet.
```
gjbex: Maybe this can be made a link as well?

source/globus/globus_platform.rst
```diff
@@ -0,0 +1,70 @@
...
+
+
+.. local_endpoints
+
+Local endpoints
```
gjbex: Would it make sense to first talk about local endpoints, and only then about file sharing?  The advantage would be that people can follow along.

source/globus/data_sharing.rst
```diff
@@ -0,0 +1,22 @@
...
+
+- Provide a name for the guest collection, and click :bgrnd1:`Create Share`.
+
+- When your collection is created, you will be taken to the Sharing tab, where you can set permissions. As shown below, the starting permissions give read and write access (and the Administrator role) to the person who created the collection. Click the :bgrnd1:`Add Permissions` button or icon to share access with others. You can add permissions for an individual user, for a group, or for all logged-in users. In the :bgrnd1:`Identity/E-mail` field, type a person's name or username (if user is selected) or a group name (if group is selected) and press :bgrnd1:`Enter`. Globus will display matching identities. Pick from the list. If the user has not used Globus before or you only have an email address, enter the email address and click :bgrnd1:`Add`. The users you share with will receive an email notification containing a link to the shared endpoint. You may add a customized message to this email. If you do not want to send a notification, uncheck the :bgrnd1:`Send E-mail` checkbox. You can add permissions to subfolders by entering a path in the :bgrnd1:`Path` field.
+
```
gjbex: Add a note warning about being careful whom to give which permissions?

source/globus/globus_platform.rst
```diff
@@ -0,0 +1,70 @@
...
+---------------
+
+.. include:: local_endpoints.rst
+
+
+.. _data_sharing
+
```
gjbex: Add a use case for Tier-2/Tier-1 data transfer?

source/globus/globus_platform.rst
avapirev: Maybe. Until the next user asks the opposite due to the interrelated nature of both.

source/globus/globus_platform.rst
avapirev: I am not sure I follow. We have this - Note At the VSC all Tier-1 and Tier-2 systems have dedicated Globus Collections. Type vsc in the search field to see the list.

source/globus/globus_platform.rst
gjbex: I'm not questioning it is technically possible and already implemented.  I'm just wondering whether it would make sense to add an explicit remark in the Globus docs that states that a very nice use case is data transfer between Tier-1 and Tier-2 across sites.  It is typically something users struggle with, and it would motivate them to start using Globus.

PR #106, opened 2019-10-15
Support request info revision

gjbex:
The guidelines for support requests have had a small update:
  * request to copy/paste terminal text rather than sending a screenshot,
  * request to add the cluster name for software that malfunctions,
  * slight rephrasing to make subsections wording consistent.

### DIFF
--- a/source/user_support.rst
+++ b/source/user_support.rst
@@ -49,6 +49,7 @@ If installed software malfunctions/crashes
 
 Please provide the following information:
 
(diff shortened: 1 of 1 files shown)


mselwa:
In all categories I would like to have vsc id not to have it checked for all the tickets.

For software - in case of running the job - location of the script or attachement of the submitted script and the same with error/output files of the job would be good to have.

Issue #122, opened 2019-11-20
Add remark that having a conda function definition in your .bashrc will give problems for NX login


gjbex:
When installing miniconda3, the installer offer to modify the user's `.bashrc`.  In the past, it simply added a modification of `PATH`, now it defines a `conda` function.

Somehow, this causes issues when users try to log in on NX nodes, so a cautionary note should be added to the documentation.

PR #123, opened 2019-11-20
Feature/conda bashrc

gjbex:
Add warnings to various pages that pertain to either conda or NX, cfr. #122

### DIFF
--- a/source/access/nx_start_guide.rst
+++ b/source/access/nx_start_guide.rst
@@ -93,7 +93,7 @@ step-by-step or to the :download:`document containing graphical
 instructions <nx_start_guide/nx_config_guide.pdf>`.
 
 How to start using NX on Genius?
-----------------------------------
+--------------------------------
 
 #. Once your desktop is open, you can use all available GUI designed
    software that is listed within the Applications menu. Software is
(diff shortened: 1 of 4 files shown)


moravveji:
- In source/access/nx_start_guide.rst, I propose an few lines like this: 

> If you would like to use NX and conda, follow these instructions for a clean setup:
>  - open your .bashrc, and find the relevant lines which modifies the $PATH and also initializes conda
>  - copy these lines and paste them into a new file called `$VSC_HOME/conda.start` 
>  - clean your .bashrc from all those lines which modifies $PATH and initializes conda
>  - You can now safely login to NX. Next time you want to use conda, you have to execute `source $VSC_HOME/conda.start` in your command prompt or inside your PBS job script

- In https://github.com/hpcleuven/VscDocumentation/blob/feature/conda_bashrc/source/leuven/running_jupyter_notebooks.rst#prerequisite-nomachine the :ref: is not rendered, because of a typo

- In https://github.com/hpcleuven/VscDocumentation/blob/feature/conda_bashrc/source/software/python_package_management.rst#install-miniconda, we are misguiding the users with adding conda path to $PATH. Instead, we can advice them to have a secondary file like `conda.start`

- The same could also apply to https://github.com/hpcleuven/VscDocumentation/blob/feature/conda_bashrc/source/software/r_package_management.rst#installing-miniconda

gjbex:
The modifications of the documentation were based on incorrect information, hence the PR is closed.

PR #131, opened 2020-03-02
Making UGent info more spoi

epauwels:


gjbex:
Please update the information on the VSC documentation site, rather than referring to local documentation.  If necessary, this will be escalated to HEC and FWO.

PR #135, opened 2020-03-24
Update tier2_hardware.rst

epauwels:
Okay, abandoning SPOI
But updated anyhow


gjbex:
@epauwels  Many thanks!

PR #137, opened 2020-04-03
Add remarks on using large amounts of memory with few cores

gjbex:


source/leuven/superdome_quick_start.rst
```diff
@@ -57,9 +57,20 @@ resources you specify are not available.
...
+   and so on.
 
 
+If your application requires fewer cores than the 14 a NUMA-node has, but
+you need more memory than the default, you should specify that explicitely.
+For instance, if you would have a sequential program that required 500 GB of
+RAM, you would use ``-L tasks=1:lprocs=1:place=numanode=1:memory=500gb``.
+
+However, keep in mind that if the memory you require exceeds that of a
+NUMA-node, you would have to request multiple nodes to accommodate for
```
moravveji, line 69: Maybe, earlier in this page, we just need to mention that each numanode has 700GB memory. 
Then, the current sentence links well with the definition.

gjbex: @moravveji Thanks, I've implemented that.

PR #142, opened 2020-04-09
Lower job walltime to comply with debug walltime limit

MaximeVdB:
An extremely tiny (but still relevant!) change, which seemed like a good occasion to start working on VscDocumentation.

So the thing is that the Breniac debug job example displayed a walltime of 2 hours, but the maximum debug queue walltime (as specified in the lines above it) is only 1 hour.


gjbex:
For Tier-1 related additions/changes, you could consider asking Ingrid for a review.  (Not necessary in this case, of course.)

PR #159, opened 2020-05-26
Feature/wnscp

mselwa:


sbecuwe:
Just one question / remark. In the current version, (numbered) boxes visualize the steps. This seems to be missing from the new version.

mselwa:
Thanks @sbecuwe. Yes, I was fighting with paint yesterday, but could not produce anything nice. Today got a brilliant idea to do that in powerpoint. For config - it is obvious, for logging - removed the numbers as they do not match with actions any more (action 1 cannot be shown on the picture).

PR #186, opened 2020-12-10
Add note on the consequence of increasing pmem 

stevenvdb:
On the KU Leuven/UHasselt Tier-2, setting pmem higher than the available memory per core, enforces a SingleJob policy.
I added a note to explain the consequences for users.

I am unsure about the policies at other sites (other than that they discourage pmem at VUB). Therefore, I am tagging some admins to see if anyone has objections to this.
@smoors @boegel @sbecuwe


gjbex (approved):

smoors:
at VUB, we now handle pmem in a submitfilter, so users are now free to use either mem or pmem.
the added note does not apply to Brussel, so this is fine for us.

source/jobs/specifying_resources.rst
```diff
@@ -165,6 +165,16 @@ In total, each of the 16 processes can use 10 GB RAM.
    users are recommended to specify the total amount of RAM requested for the job
    with ``-l mem=<total_memory>``.
 
+.. note::
+
+   On the KU Leuven/UHasselt Tier-2 cluster, setting ``pmem`` to a value
+   higher than the total amount of memory on the node divided by the number of
+   cores on the node, will lead the scheduler to reserve the entire node. This
```
MaximeVdB, line 172: How about condensing "the total amount of memory on the node divided by the number of cores on the node" into "the available memory per core"?

smoors (approved):

PR #187, opened 2021-01-26
Added 'irods-setup' command usage

mstfdkmn:
Hi Ingrid,
Could you please also review this change and merge it to data_M to be rendered? Thanks.


ingridbr (changes requested):
I miss the recommendation to use always  irods_config command when working on the computer nodes and in scripts

ingridbr (approved):

PR #203, opened 2021-05-12
remove NX protocol

jefscheepers:
The previous version suggested to use NX protocol and port 4000 if SSH gave problems.
Removed this because NX protocol has been inactivated.


gjbex:
Please revise as development will be removed.

jefscheepers:
I see this change was already implemented in the meantime. I will close the PR.

PR #205, opened 2021-05-28
add links to VNC documentation for VUB

smoors:
@gjbex to avoid duplication, maybe we should put VNC (and NX client) in a separate page and link to that?


lexming (approved):
LGTM

PR #212, opened 2021-06-17
Add note about maximal number of jobs in Genius queues

stevenvdb:
There are limits on how many jobs a user can have queued simultaneously. If a limit is exceeded, jobs end up in qdef and will not start without manual intervention. This PR adds a note about which limits are in place.

### DIFF
--- a/source/leuven/genius_quick_start.rst
+++ b/source/leuven/genius_quick_start.rst
@@ -62,6 +62,10 @@ walltimes of 3 days or less.
    - q7d: max_user_queueable = 20
    - qsuperdome: max_user_queueable = 20
 
+   These limits can be checked on the cluster by executing::
+
+      $ qstat -f -Q
+
 
 .. _submit to genius compute node:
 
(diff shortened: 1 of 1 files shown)


source/leuven/genius_quick_start.rst
```diff
@@ -48,6 +48,20 @@ with walltimes between 3 and 7 days are furthermore only allowed to request up
...
+   than the maximum number, these jobs will be deferred and will not start.
+   Therefore you should always respect the following limits on how many jobs
+   you have in a queue at the same time:
+
+   - q1h: max_user_queueable = 200
+   - q24h: max_user_queueable = 250
+   - q72h: max_user_queueable = 150
+   - q7d: max_user_queueable = 20
+   - qsuperdome: max_user_queueable = 20
+
```
MaximeVdB, line 64: Looks good -- maybe it's worth to add somewhere (e.g. just at the end) that the users can check this themselves with "qstat -f -Q"?

MaximeVdB:
Also this: better propose to merge into development, instead of the master branch.

source/leuven/genius_quick_start.rst
stevenvdb: I did this in a new commit. I hope it does not encourage users to ask for the q21d which shows up...

stevenvdb:
> 
> 
> Also this: better propose to merge into development, instead of the master branch.



> 
> 
> Also this: better propose to merge into development, instead of the master branch.

That was a mistake, fixed now.

MaximeVdB:
OK -- feel free to merge!
(or assign someone to do it)

PR #222, opened 2021-11-12
Tier1-Hortense hardware

epauwels:
Starting documentation for Tier1c Hortense

### DIFF
--- a/source/gent/tier1_hardware.rst
+++ b/source/gent/tier1_hardware.rst
@@ -1,0 +1,22 @@
+Hardware details
+----------------
+
+- partition cpu_rome
+   - 294 nodes
+   - 2x 64-core AMD Epyc 7H12 CPU 2.6 GHz
+   - 256 GiB RAM
+
+- partition cpu_rome
+   - 42 nodes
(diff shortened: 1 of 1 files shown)


source/gent/tier1_hardware.rst
```diff
@@ -0,0 +1,22 @@
+Hardware details
+----------------
+
+- partition cpu_rome
+   - 294 nodes
+   - 2x 64-core AMD Epyc 7H12 CPU 2.6 GHz
+   - 256 GiB RAM
+
+- partition cpu_rome
```
stdweird, line 9: partiton is named cpu_rome_512

epauwels: thanks, fixed now

epauwels (commented):
Comment addressed

boegel:
@epauwels Don't we want a single page with all information on Tier-1 Hortense, where there hardware info is just a subsection?
We should link to that page from https://docs.vscentrum.be/en/latest/hardware.html, and make sure it also pops up in https://docs.vscentrum.be/en/latest/

PR #237, opened 2022-02-03
Adjust Tier1 Hortense documentation

epauwels:
Added pointer to $VSC_SCRATCH_PROJECTS_BASE

### DIFF
--- a/source/gent/tier1_hortense.rst
+++ b/source/gent/tier1_hortense.rst
@@ -132,6 +132,19 @@ Keep in mind that:
 System-specific aspects
 -----------------------
 
+Project storage
+***************
+
+* Dedicated scratch storage space is reserved for all Tier1 projects.
+* The environment variable $VSC_SCRATCH_PROJECTS_BASE points to the basefolder containing all project spaces.
+* Project space is given the same name as your project.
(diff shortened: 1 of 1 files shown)


hajgato:
Maybe we have to specifically mention here, that `$VSC_SCRATCH`  is a vert limited space compared to Tier2, and it has the same purpose than `$VSC_HOME`  on Tier2.

source/gent/tier1_hortense.rst
```diff
@@ -132,6 +132,19 @@ Keep in mind that:
 System-specific aspects
 -----------------------
 
+Project storage
+***************
+
+* Dedicated scratch storage space is reserved for all Tier1 projects.
+* The environment variable $VSC_SCRATCH_PROJECTS_BASE points to the basefolder containing all project spaces.
+* Project space is given the same name as your project.
+* You can use a command of this type to directly switch to your dedicated project space:
```
itkovian, line 141: what does that mean "a command of this type"?

epauwels: better should be 'this command'

hajgato (approved):

PR #242, opened 2022-03-11
add sections on home-on-scratch + /readonly mount point to Hortense documentation

boegel:


source/gent/tier1_hortense.rst
```diff
@@ -216,28 +221,97 @@ This storage space can be addressed with the environment variable $TMPDIR
...
+
+Accessing data via ``/readonly``
+********************************
+
+Due to the fairly aggressive page cache purging policy that is used by the `Lustre <https://www.lustre.org>`_
+storage software that is used for the Tier-1 Hortense scratch filesystem, you may need to make some changes
+to how you access data in your job scripts to avoid performance problems.
+
+Whether or not this is required depends on the specific characteristics of your workloads:
+the number of files that are read, how large those files are, how the data in these files is accessed
```
stdweird, line 283: it is only relevant for data that is only read more than once. for such data, the perf impact depends on number of files, size, io pattern.

source/gent/tier1_hortense.rst
```diff
@@ -216,28 +221,97 @@ This storage space can be addressed with the environment variable $TMPDIR
...
+(the I/O pattern), etc. Note that this applies to both input data for your workloads, as well as
+any software you have installed on the Tier-1 Hortense scratch filesystem (see also :ref:`hortense_software_readonly`).
+
+To mitigate performance problems, you can access the data in your project scratch directory
+through the ``/readonly`` mount point, rather than accessing it directly.
+**We recommend to access your data via the** ``/readonly`` **mount point whenever possible,
```
stdweird, line 289: i would not write this. as it is not easy to determine this. people should try and see if it matters.
we will most likely also tune lustre further so this even less an issue (the aggressive cache flushing can be mitigated somewhat)

source/gent/tier1_hortense.rst
```diff
@@ -216,28 +221,97 @@ This storage space can be addressed with the environment variable $TMPDIR
...
+since it can result in a signficant performance speedup for your jobs.**
+
+This is done by prefixing the path to files and directories with ``/readonly/`` in your job script:
+rather than accessing your data via ``$VSC_SCRATCH_PROJECTS_BASE/...`` (or ``/dodrio/scratch/...``,
+which you should not used), you just use ``/readonly/$VSC_SCRATCH_PROJECTS_BASE/...`` instead.
```
stdweird, line 294: used -> use

source/gent/tier1_hortense.rst
```diff
@@ -216,28 +221,97 @@ This storage space can be addressed with the environment variable $TMPDIR
...
+As the name suggests, the ``/readonly`` mount point only provides *read-only* access to your data.
+Trying to make any changes to files accessing via ``/readonly`` will result in ``Permission denied`` errors.
+
+.. note::
+
+   On the login nodes, there is a delay of maximum 30 minutes for changes to files (or new/removed
+   files/directories) to be reflected through the ``/readlonly/`` mount point.
+
+   In jobs, any changes you make to files or directories in your project scratch directory should be reflected
+   through the ``/readonly`` mount point immediately, as long as the job started running *after* the
```
stdweird, line 312: the immediately is confusing, remove it.

source/gent/tier1_hortense.rst
```diff
@@ -216,28 +221,97 @@ This storage space can be addressed with the environment variable $TMPDIR
...
+   changes were made.
+
+   In addition, take into account that changes which are made while the job is running will *not* be
+   reflected through the ``/readonly`` mount point (for that job).
```
stdweird, line 316: for that -> during that

source/gent/tier1_hortense.rst
```diff
@@ -216,28 +221,97 @@ This storage space can be addressed with the environment variable $TMPDIR
...
+   If your job script creates new files, updates existing files, etc., those changes will *not* be
```
stdweird, line 317: new files/dirs might show up; but people should not assume so. not sure how to phrase this. i don;t mind making it thi strict, but it's not 100% true

source/gent/tier1_hortense.rst
```diff
@@ -216,28 +221,102 @@ This storage space can be addressed with the environment variable $TMPDIR
...
+or in case of problems with the network connection to the other VSC sites.
+
+Although your VSC home directory is usually accessible via ``$VSC_HOME``,
+we strongly recommend to *not* simply create symbolic links to files like your ``.bashrc`` startup script,
+since that would defeat the purpose of this "home-on-scratch" setup.
+
+This recommendation also applies to ``$VSC_DATA``: you should avoid using it in your job scripts as much as
+possible, and ensure that your workflow only relies on the Hortense scratch filesystem. If you require any
+data as input for your jobs, it should be copied to the Hortense scratch filesystem first (for example
+via :ref:`globus`).
```
stdweird, line 243: i would not mention only globus; it's overkill for small things and eg industry people don't have access to it

source/gent/tier1_hortense.rst
```diff
@@ -216,28 +221,100 @@ This storage space can be addressed with the environment variable $TMPDIR
...
+
+.. _hortense_project_scratch_dirs:
+
+Project scratch directories
+***************************
+
+* Please be aware that storage space on ``$VSC_SCRATCH`` (personal scratch directory) is limited per user to 3 GB.
+* Instead, it is better to use the dedicated scratch storage space which is reserved for your Tier-1 project.
+* The environment variable ``$VSC_SCRATCH_PROJECTS_BASE`` points to the base folder containing all project directories.
+* Project directories are given the same name as your Tier-1 project (so *without* a prefix like ``gpr_compute_``).
```
stdweird, line 252: @epauwels what other prefixes do we have : startin and explor?

source/gent/tier1_hortense.rst
```diff
@@ -383,6 +460,45 @@ and load one or more modules via the ``module load`` command to start using them
...
+for example in scripts or binaries that make part of the installation,
+you should install the software using the ``bwrap`` utility. This allows you to "rename" the path to your
+project scratch directory so it starts with ``/readonly/``, while preserving write access to it.
+
+Assuming that the procedure to install the software is implemented in a script named ``install.sh``,
+you can use ``bwrap`` as follows:
+
+.. code::
+
+   bwrap --bind / / --bind $VSC_SCRATCH_PROJECTS_BASE /readonly/$VSC_SCRATCH_PROJECTS_BASE --dev /dev --bind /dev/log /dev/log ./install.sh
```
stdweird, line 486: yuck. can't we make a wrapper for this?

source/gent/tier1_hortense.rst
boegel: I'm happy to fix the docs to mention all possible prefixes, but not in this PR please :)

source/gent/tier1_hortense.rst
```diff
@@ -383,6 +460,45 @@ and load one or more modules via the ``module load`` command to start using them
...
+
+The ``install.sh`` script should be implemented such that it installs the software to
```
stdweird, line 488: you can also suggest to start a shell with bwrap, and then install the software from the shell

source/gent/tier1_hortense.rst
boegel: Yeah, we should, I agree. A commando like `install-in-readonly` or something?

boegel: Or `run-in-readonly`

stdweird: in the example, i would also add the home `--bind /readonly/$HOME $HOME`

source/gent/tier1_hortense.rst
boegel: I'd say that's harder to explain... Also, we don't want to encourage people to install stuff manually.

source/gent/tier1_hortense.rst
boegel: `dodrio-bind-readonly` is now in place, which binds `/dodrio` to `/readonly/dodrio`, so no need to also bind `$HOME` (which is `/dodrio/scratch/users`)

source/gent/tier1_hortense.rst
boegel: fixed in 9ceb1e7

source/gent/tier1_hortense.rst
boegel: fixed in 9ceb1e7

stdweird (approved):

PR #255, opened 2022-05-18
Explain how automatic file removal on scratch works

stevenvdb:
We had a few (two I believe) cases where users transferred files to $VSC_SCRATCH which were nearly immediately deleted because the access timestamp was not updated. This PR clarifies how the automatic removal works and what the implications are for initially putting files on $VSC_SCRATCH.


MaximeVdB (approved):

mselwa:
Seems very clear and nicely explained, however - my idea was to add to it cleaning policy of KUL - so in the source/leuven/tier2_hardware/kuleuven_storage.rst Perhaps less affecting other universities?

mselwa:
Perfect! Thank you.

mselwa (approved):

PR #257, opened 2022-05-27
initial wICE description

jooghe:
This description is only accessible through the direct link. It can be used for the first pilot users. To make it general visible, it should also contain more reference to the slurm submit specifics.


source/leuven/tier2_hardware/wice_hardware.rst
```diff
@@ -0,0 +1,45 @@
+Genius hardware
```
gjbex, line 1: wICE hardware

gjbex (changes requested):
Enkel de titel van de pagina, verder lijkt het me prima.

PR #286, opened 2022-11-08
Update documentation workflow

gjbex:
As the development branch will be deleted, the workflow description has been updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,15 +39,13 @@ $ source activate sphinx
 
 ### Creating a feature branch
 
-Do not make changes in the master or development branch directly, but create feature/bugfix branches as required, e.g.,
+Do not make changes in the master branch directly, but create feature/bugfix branches as required, e.g.,
 ```bash
-$ git checkout development
-$ git checkout -b feature/new_stuff
+$ git switch master
+$ git switch -c feature/new_stuff
 ```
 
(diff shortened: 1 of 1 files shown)


smoors (approved):
lgtm

backelj (approved):

README.md
```diff
@@ -39,15 +39,13 @@ $ source activate sphinx
...
-$ git checkout development
-$ git checkout -b feature/new_stuff
+$ git switch master
+$ git switch -c feature/new_stuff
 ```
 
-Note that at any given time, we should be able to merge the development branch into the master branch, and rebuild the documentation, hence the development branch should always be in a sane state.
-
-Exception on feature/bugfix branches: use common sense, to fix a simple typo, it is likely better to do that directly in the development branch, and merge that immediately into master.
+Exception on feature/bugfix branches: use common sense, to fix a simple typo, it is likely better to do that directly in the master branch.
```
moravveji, line 48: as a matter of fact, numerous small typo fixes directly done on the master branch led to large divergence of the two branches. So, I would even drop this line and make no exceptions (because exceptions are left to subjective interpretations)  ;-)

moravveji (changes requested):

README.md
gjbex: @moravveji Well, no, the discrepancies started to occur because some people blatantly ignored the repositories policies, which is bad form.

Regardless, I removed the remark on "common sense", quite appropriately in this context.

moravveji (approved):

PR #291, opened 2023-01-02
Slurm accounting

mselwa:


source/leuven/wice_quick_start.rst
```diff
@@ -27,7 +27,8 @@ In these examples, the submission will be done from login nodes of the Genius cl
 To submit to a compute node you need to provide the required number of nodes and cores. For example to request 2 nodes with each 72 cores for 2 hours you can submit like this::
 
    $ sbatch --cluster=wice --nodes=2 --ntasks-per-node=72 --time=2:00:00  -A lp_myproject  myjobscript.slurm
-
+   
+ More information about accounting used on wice can be found :ref:`here <_accounting_leuven>`.
```
stevenvdb, line 31: * You need to remove the space at the front, currently this line is included in the code box from the lines above.
* The link should be `<accounting_leuven>` instead of `<_accounting_leuven>`.

stevenvdb (changes requested):
I added a few minor comments/suggestions in the text.

One bigger comment is the question whether we should say something about the coexistence of Slurm and MAM credits (which will be temporary if I am not mistaken). Some things that are not clear to me:
* Should users contact us if they want to transfer credits between a MAM and a Slurm account or can they (project leaders) do it themselves?
* Should we discuss the conversion rate?
* Should all of this all go (only) in the Service Catalog?

mselwa:
> I added a few minor comments/suggestions in the text.
> 
> One bigger comment is the question whether we should say something about the coexistence of Slurm and MAM credits (which will be temporary if I am not mistaken). Some things that are not clear to me:
> 
>     * Should users contact us if they want to transfer credits between a MAM and a Slurm account or can they (project leaders) do it themselves?

I believe that will be in a service catalog. The users need to contact us as mam-commands are executed only by the admins (hopefully slurm as well is as it used to be - by root)
> 
>     * Should we discuss the conversion rate?

I tried to get the info from Jan - when stearing group decided to lower them, but got a very enigmatic reply. I believe it will be finally given in SC? If not - once Peter arranges a new formula we can add it. The olds one will be wrong anyway.
> 
>     * Should all of this all go (only) in the Service Catalog?
That was also in discussion with KAOS - who has write rights on the file. I do not see changes yet, so perhaps they were not published or not finalized...

stevenvdb:
> > I added a few minor comments/suggestions in the text.
> > One bigger comment is the question whether we should say something about the coexistence of Slurm and MAM credits (which will be temporary if I am not mistaken). Some things that are not clear to me:
> > ```
> > * Should users contact us if they want to transfer credits between a MAM and a Slurm account or can they (project leaders) do it themselves?
> > ```
> 
> I believe that will be in a service catalog. The users need to contact us as mam-commands are executed only by the admins (hopefully slurm as well is as it used to be - by root)
> 
> > ```
> > * Should we discuss the conversion rate?
> > ```
> 
> I tried to get the info from Jan - when stearing group decided to lower them, but got a very enigmatic reply. I believe it will be finally given in SC? If not - once Peter arranges a new formula we can add it. The olds one will be wrong anyway.
> 
> > ```
> > * Should all of this all go (only) in the Service Catalog?
> > ```
> 
> That was also in discussion with KAOS - who has write rights on the file. I do not see changes yet, so perhaps they were not published or not finalized...

I was not closely involved in this, but keeping everything in the Service Catalog seems like a good idea to me. Perhaps add a link from here?

Another point that might be interesting to mention here, is that users can compare the credit cost of different partitions by executing ``scontrol --cluster=wice show partition`` and looking for the ``TRESBillingWeights`` field, which gives the number of credits per core-minute (or per GPU-minute).

source/leuven/slurm-accounting.rst
```diff
@@ -0,0 +1,56 @@
+.. _accounting_leuven:
+========================
+Slurm Accounting
+========================
+
+:ref:`wICE <wice_t2_leuven>` cluster uses Slurm Account Manager for credits accounting purpose. We developed so called SAM-commands to mimic the behavior of previously used MAM (moab account manager). From the user point of view it should be enough to change ``mam`` phrase in the accounting command into ``sam``.
```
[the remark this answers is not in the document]
moravveji: ... Slurm Account Manager (SAM) ...

moab -> MOAB

source/leuven/slurm-accounting.rst
```diff
@@ -0,0 +1,56 @@
...
+
+   $ sam-statement  -A lp_my_project  -s 2023-01-01  -e 2023-01-31
+
+This will show the transactions on the account for the ``lp_my_project`` project for the month january 2023.
+
+If you are only interested in the overview of user's transactions, and don't require the actual jobs and balance information, ``sam-list-usagerecords`` provides a much faster alternative for summarized statement::
+
+   $ sam-list-usagerecords  -A lp_my_project  -s 2023-01-01  -e 2023-01-31
+
+.. note::
```
moravveji, line 52: I think these remarks can be formatted as a bullet-point list.

Please also add that all sam-commands generate a help message, which can be inquired by providing `-h` option.

moravveji (approved):
I think this page properly reflects the current state of the sam-commands and their intent. 
Personally, I think we can leave off discussing credit conversions and rates here. The former has not happened yet, and the latter must be differed to service catalog (with a link as steven proposed).

PR #294, opened 2023-01-06
 Python SDK documentation + small fixes

jefscheepers:


source/globus/cli.rst
```diff
@@ -58,15 +58,18 @@ You can find the names of all VSC collections in our table of :ref:`globus-avail
...
-      $ globus ls $tier2scratch        
+      $ globus ls $kuleuventier2scratch        
             nproject/
             nscratch/
             project/
             scratch/                                                            
       
       # To list a specific directory:
-      $ globus ls $tier2scratch:scratch/337/vsc33731/test
+      $ globus ls $kuleuventier2scratch:scratch/<redacted>/<username>/test
```
ingridbr, line 72: What this "redacted" here means? The number that is used on the path? It is this explained somewhere???

source/globus/python_sdk.rst
```diff
@@ -0,0 +1,234 @@
...
+      transfer_client = globus_sdk.TransferClient(authorizer=authorizer)
+
+After that, we create a transfer task::
+
+      task_data = globus_sdk.TransferData(
+            # we specify the transfer client, source collection and destination collecion
+            transfer_client, VUB_tier2, KULeuven_tier2_scratch
+      )
+
+Then, we add data to the transfer::
```
ingridbr, line 176: This code will work if you are transferring 1 single file. For transferring a directory between 2 endpoints you will need to add in the add_item the parameter recursive=True or the transfer will fail with and error. For more information see: 

https://globus-sdk-python.readthedocs.io/en/stable/services/transfer.html#globus_sdk.TransferData.add_item

I think this should be explain here

ingridbr (changes requested):

source/globus/cli.rst
jefscheepers: Indeed unclear, changed it to 'xxx' and 'vscxxxxx' and added a comment about it.

source/globus/python_sdk.rst
jefscheepers: Done in my last commit

mstfdkmn (approved):

PR #297, opened 2023-01-11
Feature/matlab parallel

w-lampaert:


source/software/matlab_parallel_computing.rst
```diff
@@ -3,12 +3,126 @@
...
+Using parallel computing together with Matlab does not work out of the box. You will need to follow some (simple) configurations steps, that will be explained here.
+
+First of all, you will need some scripts that will configure Matlab for parallel use. You can download the appropriate script from this 
+`repository <https://github.com/hpcleuven/matlab-remote>`_. Place this folder somewhere on your cluster storage (e.g. ``$VSC_DATA``). Set your working directory to
+the directory where you saved the scripts. Within the folder, you will see a ``mdcs.rc`` file. This is a configuration file that allows you to adapt the cluster
+profile to your needs. There is only one setting that needs to be changed with the approach you are using here: ``LocalJobStorageLocation``. This will define where
+your job output will be directed to. There is no need for quotation marks when defining this path. **Be careful**: when leaving this field blank, this location will 
+default to your ``$VSC_HOME``, which could result in quickly filling up your home directory.
+
+Next, you'll need to load the module, and start it. It is recommended to request an interactive job for this beforehand.
```
MaximeVdB, line 31:
```suggestion
Next, you'll need to load the module, and start it. It is recommended to do this via an interactive job.
```
(easier to read)

MaximeVdB (changes requested):

source/software/matlab_parallel_computing.rst
```diff
@@ -3,12 +3,126 @@
...
+
+   >> % View the additional properties
+   >> c.AdditionalProperties
+   >> % set the required properties
+   >> c.AdditionalProperties.AccountName = <account-name>;
+   >> c.AdditionalProperties.WallTime = '05:00:00';
+   >> % save the profile
+   >> c.saveProfile
+   
+Now the profile is ready, and you should be able to submit multi-node jobs with Matlab now.
```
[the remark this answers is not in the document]
w-lampaert: The explanation of this was a bit above (line 63). I moved it down because it does indeed make more sense to put it here.

WouterVanAssche (approved):

source/software/matlab_parallel_computing.rst
```diff
@@ -3,12 +3,126 @@
...
--  The `Parallel Computing
-   Toolbox <https://nl.mathworks.com/products/parallel-computing.html>`_
-   is a regular Matlab toolbox that lets you write parallel Matlab
-   applications or use parallel implementations of algorithms in other
-   toolboxes.
-   Try ``help distcomp`` to see if the toolbox is installed for the
-   version of Matlab that you're using.
+Matlab allows you to use parallel computing on a HPC cluster using the `Parallel Computing Toolbox <https://www.mathworks.com/products/parallel-computing.html>`_.
+Only Matlab installations that have this installed, can be configured to use this. This documentation will describe how to configure this. Be aware that this has only
+been tested for some installations on the KU Leuven Tier-2 clusters.
```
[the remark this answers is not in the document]
w-lampaert: The main issue here is that the scripts I have are created for Genius specifically. So while I expect that other sites might have a similar approach, I cannot say for sure that this will actually be the case. So for now it would seem that the general part would be only the info of the parallel computing toolbox. I can of course add an extra note that states that this info is only applicable to Genius for the time being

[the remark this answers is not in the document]
MaximeVdB: OK, but even in that case I think it's good to separate the two,
i.e. to keep that one parallel computing toolbox paragraph in
a first "general" section and then put all the rest in a KU Leuven
section (which then gets further divided in subsections etc.)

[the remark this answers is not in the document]
w-lampaert: I have made some adaptations, should be oknow

source/software/matlab_parallel_computing.rst
```diff
@@ -3,12 +3,131 @@
...
+   >> c.AdditionalProperties.WallTime = '05:00:00';
+   >> % save the profile
+   >> c.saveProfile
+   
+Now the profile is ready, and you should be able to submit multi-node jobs with Matlab now.
+
+Submitting jobs
++++++++++++++++
+
+You can start both interactive and independent batch jobs with this profile. Here, we will show how to start these jobs from an opened Matlab session.
```
MaximeVdB, line 92: Do I understand it correctly that Matlab submits these jobs for you?
And so that you always need to first launch Matlab in some way, on a login node,
and that those commands below will create/submit/process jobs?
If so, it would be good to mention this more explicitly, because this would be
quite different from standard batch jobs (where you just write a batch script
and then submit that batch script with qsub).

w-lampaert: You indeed need to do this from within Matlab. I will modify the explanation

MaximeVdB (changes requested):

source/software/matlab_parallel_computing.rst
```diff
@@ -3,12 +3,136 @@
...
+loaded profile.
+
+::
+
+   >> % get a handle on the cluster
+   >> c = parcluster;
+   >> % If you would prefer to use the local resources (i.e. the resources of the node you are currently on), use:
+   >> c = parcluster('local');
+
+If you would have multiple cluster profiles, you can also load the correct cluster profile with:
```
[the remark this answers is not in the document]
w-lampaert: What do you mean with "saving the profile"? As soon as you configure a profile for a cluster-module combo, it will be saved with a standard name "<cluster_name> <matlab_version>". The "local" profile is the profile for when you want to use the resources from the node you are on. 

line 74 illustrates how to load a specific profile, but I am actually thinking of removing this altogether. You can use e.g. 'c=parcluster("Genius R2022a")' instead of using c=parcluster (of course when Genius R2022a is loaded), but there is no reason for it. Matlab will automatically detect which profile it needs to use. So using full profile names can only lead to mistakes.

source/software/matlab_parallel_computing.rst
```diff
@@ -3,12 +3,136 @@
...
+   >> c.saveProfile
+   
+Now the profile is ready, and you should be able to submit multi-node jobs with Matlab now.
+
+Submitting jobs
++++++++++++++++
+
+Submitting jobs will happen from within Matlab. The general approach to launch any job in such a way, would be to start an interactive job, load your preferred
+Matlab module and start Matlab. You can do this immediately in the terminal by using the ``-nodisplay´´ flag, but you can also launch a GUI by starting Matlab on
+NX or a compute node with a GPU (it is also possible to start the GUI on nodes without a GPU, but be aware that this might hurt performance significantly).
```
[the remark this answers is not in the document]
w-lampaert: This is indeed confusing. I'll make some changes here

source/software/matlab_parallel_computing.rst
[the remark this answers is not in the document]
MaximeVdB: So do you mean that you can only have 1 profile for a specific cluster-Matlabversion combination (in addition to what you could call the "local" cluster profile)? If so, it would indeed be good to make this clear.

source/software/matlab_parallel_computing.rst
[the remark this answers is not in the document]
w-lampaert: There was a mistake here. It is not possible to submit jobs from the login node. You need to be on a compute node.

source/software/matlab_parallel_computing.rst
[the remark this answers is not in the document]
w-lampaert: I have changed extended the explanation starting from 'Configuring Matlab for parallel computing"

source/software/matlab_parallel_computing.rst
```diff
@@ -3,12 +3,155 @@
...
+   >> configCluster;
+   
+You will get a message as follows if it succeeded: 'Complete.  Default cluster profile set to "<cluster_name> <matlab_version>".'. You have now created a cluster
+profile.
+
+Configuring additional properties
+*********************************
+
+It is possible to set some user-specific properties and save them in your cluster profile. You can always adapt these and save the profile again, without performing
+the previous step again. All of these properties are available in the ``AdditionalProperties´´ property of your cluster profile.
```
MaximeVdB, line 76:
```suggestion
the previous step again. All of these properties are available in the ``AdditionalProperties`` property of your cluster profile. 
```

source/software/matlab_parallel_computing.rst
```diff
@@ -3,12 +3,155 @@
...
+
+Interactive job
+***************
+
+You can start an interactive job using the ``parpool`` function:
+
+::
+
+    >> c = parcluster;
+    >> p = parpool(64); % requesting 64 cores
```
MaximeVdB, line 126: Could you just add one or two more lines about what is supposed to be happening now? I.e.:
* When the `parpool` call returns, does that mean that your interactive job has started to run?
* If you then want to execute something in that interactive job (and not in the parent interactive
  Matlab prompt), what do you then need to do?

(The batch job example is fairly clear about this, but for the interactive case this can be made clearer.)

source/software/matlab_parallel_computing.rst
```diff
@@ -3,12 +3,155 @@
...
+Using parallel computing together with Matlab does not work out of the box. Matlab needs a 'cluster profile', specifically for the 'cluster-matlab version' combination
+you are planning to use. This cluster profile contains information like the amount of nodes, the resource manager, the way to submit jobs on that cluster... 
+Configuring these settings can be done by running the correct script once. Next to this, there are some user-specific options that can be defined. 
+
+Configuring a cluster profile
+*****************************
+
+We have provided some scripts that will allow to easily set up such a profile. You can download them from this 
+`repository <https://github.com/hpcleuven/matlab-remote>`_. Clone this folder somewhere on your personal storage (e.g. ``$VSC_DATA``). Set your working directory to
+the the cloned repository:
```
MaximeVdB, line 36:
```suggestion
the cloned repository:
```

source/software/matlab_parallel_computing.rst
```diff
@@ -3,12 +3,155 @@
...
+   >> c.saveProfile
+   
+Now the profile is ready, and you should be able to submit multi-node jobs with Matlab now.
+
+Submitting jobs
++++++++++++++++
+
+Submitting jobs will happen from within Matlab. First of all, you will need to start an interactive session on the cluster. It is possible to use the GUI, but then
+you will need to request GPU resources. You can also open the GUI on NX, but you need to submit your jobs from a compute node. Submitting from a login node
+will cause an error.
```
MaximeVdB, line 113: I think this also needs some more polishing. In particular the sentence ``You can also open the GUI on NX, but you need to submit your jobs from a compute node.`` seems contradictory. Maybe it's best to just first describe the different possible for the GUI approach, and after that describing the CLI approach (and making clear in both cases on which kind of nodes one's Matlab prompt is supposed to be running for it to work).

w-lampaert: Yes, I was not sure about how much detail I need to add here. I assume that most Matlab users are aware that they can open the GUI on NX, but like I mentioned, you cannot submit multi-node jobsfrom there. So if you want to be able to submit multi-node jobs AND use a GUI, you should request GPU resources. But, I'm not really keen on giving too much detail on that, as I do not want to push users to out GPU resources if it is just for running a GUI.

MaximeVdB: Yes, then you can indeed put most emphasis on the CLI approach. If you don't want to describe the full situation for the GUI, then you can focus on the specific case of using-the-GUI-inside-a-NX-session and then add a note that this will not allow for multi-node jobs.

source/software/matlab_parallel_computing.rst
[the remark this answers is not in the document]
w-lampaert: I have to come back on my words. This was due to an unknown issue that temporarily did not allow me to submit jobs from the login node.

source/software/matlab_parallel_computing.rst
```diff
@@ -3,12 +3,191 @@
...
+   
+Now the profile is ready, and you should be able to submit multi-node jobs with Matlab now.
+
+Submitting jobs
++++++++++++++++
+
+Submitting jobs will happen from within Matlab. There are two ways to do this. You can start a Matlab session on a login node, or you can first request an
+interactive job, where you launch your Matlab session. Which one you choose will depend on the workload you plan to have on the node before submitting a job.
+As for any other work on the cluster, you can always do minor work on the login node, but for heavier calculations (before submitting your multinode job), you should
+use an interactive session.
```
MaximeVdB, line 114:
```suggestion
use an interactive job. 
```
("session" is a bit ambiguous, especially in a context mentioning (interactive) Matlab sessions)

source/software/matlab_parallel_computing.rst
```diff
@@ -3,12 +3,191 @@
...
+Submitting jobs
++++++++++++++++
+
+Submitting jobs will happen from within Matlab. There are two ways to do this. You can start a Matlab session on a login node, or you can first request an
+interactive job, where you launch your Matlab session. Which one you choose will depend on the workload you plan to have on the node before submitting a job.
+As for any other work on the cluster, you can always do minor work on the login node, but for heavier calculations (before submitting your multinode job), you should
+use an interactive session. 
+
+It is possible to use the Matlab GUI as well, on all login nodes. It is strongly recommended though to use NoMachine if you want to use the GUI though (login node
+3 and 4), as the starting the GUI from any other login node will be extremely slow. If you want to use a GUI on a compute node, you would need GPU resources
```
MaximeVdB, line 117:
```suggestion
3 and 4), as starting the GUI from any other login node will be extremely slow. If you want to use a GUI on a compute node, you would need GPU resources
```

source/software/matlab_parallel_computing.rst
```diff
@@ -3,12 +3,191 @@
...
+(no X11 forwarding on the other nodes). Requesting GPU resources just for the GUI is quite expensive though. Once the Matlab Parallel Computing Toolbox is
```
MaximeVdB, line 118: I don't think this is true (i.e. there is no need to request GPU resources in order to use the Matlab GUI in an interactive job on a compute node). Of course, to be sure that X forwarding is enabled one needs to use the `-X` qsub option. I checked this on a Skylake debug node and it worked fine with the `matlab/R2019a` module (both via jobs-submitted-via-NoMachine and via-MobaXterm). In general the NoMachine way should be more appropriate for users on a comparatively slow internet connection (in those cases the "plain-SSH-connection" way will be a bit laggy). Could you verify this with e.g. `MATLAB/R2022b` to see if this is module-dependent?

MaximeVdB (changes requested):

source/software/matlab_parallel_computing.rst
w-lampaert: Indeed, my bad. I had issues with the x forwarding in my previous tests for some reason. I am adding that now. On my side this is still quite laggy though, especially using anything 'clickable'

source/software/matlab_parallel_computing.rst
```diff
@@ -3,12 +3,190 @@
...
+        '-l nodes=2:ppn=32 -l pmem=4gb -A '<account_name>' -l walltime=00:30:00'
+
+    Connected to the parallel pool (number of workers: 64).
+
+    ans =
+
+        ClusterPool with properties:
+
+                Connected: true
+               NumWorkers: 64
```
MaximeVdB, line 149: Using only 32 cores per node is something we don't generally recommend on Genius, so I think it's better to change the example to use 36 cores per node:
```suggestion
    >> p = parpool(72); % requesting 72 cores
    
Once the job has started, you'll receive output like this:

::

    Starting parallel pool (parpool) using the 'genius R2022a' profile ...

    additionalSubmitArgs =

        '-l nodes=2:ppn=36 -l pmem=4gb -A '<account_name>' -l walltime=00:30:00'

    Connected to the parallel pool (number of workers: 72).

    ans =

        ClusterPool with properties:

                Connected: true
               NumWorkers: 72
```

source/software/matlab_parallel_computing.rst
```diff
@@ -3,12 +3,190 @@
...
++++++++++++++++
+
+Submitting jobs will happen from within Matlab. There are two ways to do this. You can start a Matlab session on a login node, or you can first request an
+interactive job, where you launch your Matlab session. Which one you choose will depend on the workload you plan to have on the node before submitting a job.
+As for any other work on the cluster, you can always do minor work on the login node, but for heavier calculations (before submitting your multinode job), you should
+use an interactive job. 
+
+It is possible to use the Matlab GUI as well, on all login nodes. It is strongly recommended though to use NoMachine if you want to use the GUI though (login node
+3 and 4), as starting the GUI from any other login node will be extremely slow. If you want to use a GUI on a compute node, you'd need to submit your interacive job
+with X forwarding. You'll need to ``-X`` to your qsub command. Note that this is also quite slow.
```
MaximeVdB, line 118:
```suggestion
In general, we recommend to carry out the examples below via Matlab's command line interface
(as shown in the configuration section). It is also possible to use the Matlab GUI, in which case it
is best to do this via a NoMachine connection. If you then need to launch the Matlab GUI in an
interactive job, don't forget to add ``-X`` option to qsub and keep in mind that the GUI will
necessarily be less responsive.
```

MaximeVdB (changes requested):
Looks good now! Just these two last minor ones.

MaximeVdB (approved):

PR #298, opened 2023-01-12
Update genius_quick_start.rst

moravveji:
This brief PR (which must have had happened long long ago) updates the node policy on Genius.

### DIFF
--- a/source/leuven/genius_quick_start.rst
+++ b/source/leuven/genius_quick_start.rst
@@ -83,23 +83,13 @@ For instance, to test a multi-threaded application which performs optimally usin
    
 In the two above examples, the jobs may start on Skylake or Cascadelake nodes.
 Please bear in mind to not exceed the maximum allowed resources on compute nodes for the targeted partition.
-E.g. you can request at most 36 cores per node (``ppn=36``).
+E.g. you can request at most 36 cores per node (``ppn=36``). In general, we advise you to only request as much resources as needed by your application.
 
 Advanced node usage
 ^^^^^^^^^^^^^^^^^^^
-The node access policy on Skylake nodes is ``SINGLEUSER``.
-This means that once a job lands on a Skylake node(s), no job from other users can land on the same node(s).
-If you insist on using a full node (to exclude jobs from other users), you may enforce getting a Skylake node in one of the following ways::
+In certain cases (such as performance tests) you may want to be sure that your job runs on a specific type of node (i.e. only Skylake nodes or only Cascadelake nodes). You can do this by selecting a node feature via e.g. `-l nodes=1:ppn:8:skylake` or `-l feature=skylake` (same for `cascadelake`).
 
-   $ qsub -l nodes=1:ppn=8:skylake -l walltime=30:00 -A myproject myjobscript.pbs             # or
-   $ qsub -l nodes=1:ppn=8 -l feature=skylake -l walltime=30:00 -A myproject myjobscript.pbs
-
-The node access policy on Cascadelake nodes is ``SHARED``.
-This means the CPU and memory resources per nodes are exploited as much as possible by packing more and more jobs into a single node.
-Similarly, you may enforce getting a Cascadelake node by specifying either ``nodes=1:ppn=8:cascadelake`` or ``-l feature=cascadelake``
-when only needing 8 cores.
-The ``SHARED`` node access policy leaves room for smaller jobs to start executing earlier than initially scheduled.
-Therefore, users are adviced to request only as much resources as needed by their applications.
+When doing so, you should take into account that all jobs on the Skylake nodes are subjected to the `SINGLEUSER` node access policy. This means that once a Skylake node is allocated to a job, no job from other users can land on this node (even if the original job only requested a small part of the node's resources). This is different on the Cascadelake nodes, where small jobs (less than 18 cores and with default memory requirements) are given the `SHARED` node access policy instead. This allows multiple small jobs from different users to run on the same node.
 
 .. _submit to genius GPU node:
 
--- a/source/leuven/genius_quick_start.rst
+++ b/source/leuven/genius_quick_start.rst
@@ -77,10 +77,15 @@ This means that once a job lands on a Skylake node, no job from other users can 
 You may e.g. request two full nodes like this:
 
    $ qsub -l nodes=2:ppn=36  -l walltime=2:00:00  -A myproject  myjobscript.pbs
+   
+If you insist on using a full node, you may enforce getting a Skylake node in one of the following ways
+
+   $ qsub -l nodes=1:ppn=8:skylake -l walltime=30:00 -A myproject myjobscript.pbs # or
+   $ qsub -l nodes=1:ppn=8 -l feature=skylake -l walltime=30:00 -A myproject myjobscript.pbs
 
 The node access policy on Cascadelake nodes is 'shared'.
 This means the CPU and memory resources per nodes are exploited as much as possible by packing more and more jobs into a single node.
-The 'shared' node access policy leaves room for smaller jobs to hit the queue and start earlier than scheduled initially.
+The 'shared' node access policy leaves room for smaller jobs to start executing earlier than initially scheduled.
 Therefore, users are adviced to request only as much resources as needed by their applications.
 
 For instance, to test a multi-threaded application which performs optimally using 4 cores, you may submit your job like this:
--- a/source/leuven/genius_quick_start.rst
+++ b/source/leuven/genius_quick_start.rst
@@ -71,9 +71,18 @@ walltimes of 3 days or less.
 
 Submit to a compute node
 ~~~~~~~~~~~~~~~~~~~~~~~~
-To submit to a compute node it all boils down to specifying the required number of nodes and cores. As the nodes have a single user policy we recommend to always request all available cores per node (36 cores in this case). For example to request 2 nodes with each 36 cores you can submit like this::
+Submitting a compute job boils down to specifying the required number of nodes, cores-per-node, memory and walltime.
+The nodes on the ``pbs``, ``bigmem`` and ``gpu`` partitions have ``SHARED`` policy (but the nodes on the ``amd`` partition have ``SINGLEUSER`` policy).
+This means the CPU and memory resources per nodes are exploited as much as possible by packing more and more jobs into a single node.
+The ``SHARED`` node policy leaves room for smaller jobs to hit the queue and start earlier than scheduled initially.
+Therefore, users are adviced to request only as much resources as needed by their applications.
 
-   $ qsub -l nodes=2:ppn=36  -l walltime=2:00:00  -A myproject  myjobscript.pbs
+For instance, to test a multi-threaded application which performs optimally using 4 cores, you may submit your job like this:
+
+   $ qsub -l nodes=1:ppn=4  -l walltime=2:00:00  -A myproject  myjobscript.pbs
+   
+However, make sure that you do not exceed the maximum allowed resources on compute nodes for the targeted partition.
+E.g. you can use maximum 36 cores per node ``ppn=36``.
   
 
 .. _submit to genius GPU node:
--- a/source/leuven/genius_quick_start.rst
+++ b/source/leuven/genius_quick_start.rst
@@ -87,9 +87,9 @@ E.g. you can request at most 36 cores per node (``ppn=36``). In general, we advi
 
 Advanced node usage
 ^^^^^^^^^^^^^^^^^^^
-In certain cases (such as performance tests) you may want to be sure that your job runs on a specific type of node (i.e. only Skylake nodes or only Cascadelake nodes). You can do this by selecting a node feature via e.g. `-l nodes=1:ppn:8:skylake` or `-l feature=skylake` (same for `cascadelake`).
+In certain cases (such as performance tests) you may want to be sure that your job runs on a specific type of node (i.e. only Skylake nodes or only Cascadelake nodes). You can do this by selecting a node feature via e.g. ``-l nodes=1:ppn:8:skylake`` or ``-l feature=skylake`` (same for ``cascadelake``).
 
-When doing so, you should take into account that all jobs on the Skylake nodes are subjected to the `SINGLEUSER` node access policy. This means that once a Skylake node is allocated to a job, no job from other users can land on this node (even if the original job only requested a small part of the node's resources). This is different on the Cascadelake nodes, where small jobs (less than 18 cores and with default memory requirements) are given the `SHARED` node access policy instead. This allows multiple small jobs from different users to run on the same node.
+When doing so, you should take into account that all jobs on the Skylake nodes are subjected to the ``SINGLEUSER`` node access policy. This means that once a Skylake node is allocated to a job, no job from other users can land on this node (even if the original job only requested a small part of the node's resources). This is different on the Cascadelake nodes, where small jobs (less than 18 cores and with default memory requirements) are given the ``SHARED`` node access policy instead. This allows multiple small jobs from different users to run on the same node.
 
 .. _submit to genius GPU node:
 
@@ -140,7 +140,7 @@ The big memory nodes are also located in a separate partition. In case of the bi
 
 Submit to an AMD node
 ~~~~~~~~~~~~~~~~~~~~~
-All jobs on AMD nodes are given the `SINGLEUSER` node access policy
+All jobs on AMD nodes are given the ``SINGLEUSER`` node access policy
 (see above "Advanced node usage" paragraph for more information).
 Besides specifying the partition, it is also important to specify the memory 
 per process (``pmem``) since the AMD nodes have 256 GB of RAM, which implies 
--- a/source/leuven/genius_quick_start.rst
+++ b/source/leuven/genius_quick_start.rst
@@ -83,7 +83,7 @@ For instance, to test a multi-threaded application which performs optimally usin
    
 In the two above examples, the jobs may start on Skylake or Cascadelake nodes.
 Please bear in mind to not exceed the maximum allowed resources on compute nodes for the targeted partition.
-E.g. you can use maximum 36 cores per node ``ppn=36``.
+E.g. you can request at most 36 cores per node (``ppn=36``).
   
 
 .. _submit to genius GPU node:
--- a/source/leuven/genius_quick_start.rst
+++ b/source/leuven/genius_quick_start.rst
@@ -140,7 +140,8 @@ The big memory nodes are also located in a separate partition. In case of the bi
 
 Submit to an AMD node
 ~~~~~~~~~~~~~~~~~~~~~
-The AMD nodes are in their own partition, with `SINGLEUSER` node access policy.
+All jobs on AMD nodes are given the `SINGLEUSER` node access policy
+(see above "Advanced node usage" paragraph for more information).
 Besides specifying the partition, it is also important to specify the memory 
 per process (``pmem``) since the AMD nodes have 256 GB of RAM, which implies 
 that the default value is too high, and your job will never run.
--- a/source/leuven/genius_quick_start.rst
+++ b/source/leuven/genius_quick_start.rst
@@ -84,7 +84,22 @@ For instance, to test a multi-threaded application which performs optimally usin
 In the two above examples, the jobs may start on Skylake or Cascadelake nodes.
 Please bear in mind to not exceed the maximum allowed resources on compute nodes for the targeted partition.
 E.g. you can request at most 36 cores per node (``ppn=36``).
-  
+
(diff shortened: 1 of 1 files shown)


mselwa (approved):
I find it ok.

source/leuven/genius_quick_start.rst
```diff
@@ -71,9 +71,18 @@ walltimes of 3 days or less.
 
 Submit to a compute node
 ~~~~~~~~~~~~~~~~~~~~~~~~
-To submit to a compute node it all boils down to specifying the required number of nodes and cores. As the nodes have a single user policy we recommend to always request all available cores per node (36 cores in this case). For example to request 2 nodes with each 36 cores you can submit like this::
-
-   $ qsub -l nodes=2:ppn=36  -l walltime=2:00:00  -A myproject  myjobscript.pbs
+Submitting a compute job boils down to specifying the required number of nodes, cores-per-node, memory and walltime.
+The nodes on the ``pbs``, ``bigmem`` and ``gpu`` partitions have ``SHARED`` policy (but the nodes on the ``amd`` partition have ``SINGLEUSER`` policy).
```
MaximeVdB, line 75: It looks like you're talking about Moab's "node allocation policy",
which is how Moab is deciding which node(s) to allocate for a given job.
The policy employed on Genius seems to be the "MINRESOURCE" one,
where indeed eligible small jobs are preferentially sent to nodes which
are already partially being used by other small jobs.

I think it would be best if you would rephrase this accordingly, and so avoiding
terms like "SHARED" and "SINGLEUSER" and writing "node allocation policy"
instead of "node policy". This would reduce the confusion with what is called
"node access policy" (which is a related topic, but not one that needs to be
addressed in this Quickstart page).

Also, so far I couldn't find out whether the node allocation policy is 
different on the AMD partition compared to the other partitions.
As far as I can see, it's "MINRESOURCE" everywhere.

mselwa:
Ok. I thought it has been confirmed what was changed, But apparently - sit really. cascadelakes are shared, but skylakes did not change
 $ checknode r22i13n01
node r22i13n01

State:   Running  (in current state for 00:12:44)
Configured Resources: PROCS: 36  MEM: 184G  SWAP: 185G  DISK: 1M  socket: 2  numanode: 2  core: 36  thread: 36
Utilized   Resources: PROCS: 18
Dedicated  Resources: PROCS: 18  MEM: 90G
Attributes:         kernel=3-10-0-1160-81-1-el7-x86_64,puppethpc=20230110-09-49-143-15git52e6367,puppethpccode=20230111-6124-git6a9404cf,sudo=1-8-23-10-el7_9-2
  MTBF(longterm): 7:21:42:00  MTBF(24h):   INFINITY
Opsys:      linux     Arch:      ---
Speed:      1.00      CPULoad:  17.180
Partition:  pbs  Rack/Slot:  22/12  NodeIndex:  22
Features:   singleisland,beeond,skylake,r22i13,r22,type_skylake,genius,mem192
NodeType:   skylake
Classes:    [q1h][q24h][q72h][q7d][q21d][qsuperdome]
RM[pbs]*:   TYPE=PBS  STATE=Running
EffNodeAccessPolicy: SINGLEJOB

Total Time:   INFINITY  Up:   INFINITY (97.74%)  Active:    613days (34.85%)

Reservations:
  debug.15178x1  User    -979days ->   INFINITY (  INFINITY)
    Blocked Resources@  -979days  Procs: 36/36 (100.00%)  Mem: 0/188492 (0.00%)  Swap: 0/189985 (0.00%)  Disk: 0/1 (0.00%)
  51372091x1  Job:Running  -00:13:28 -> 00:16:31 (00:29:59)
Jobs:        51372091

✔ [Jan/12 15:44] x0076109@tier2-p-login-3 ~ $ checknode r26i13n01
node r26i13n01

State:  Draining  (in current state for 00:00:00)
Configured Resources: PROCS: 36  MEM: 184G  SWAP: 185G  DISK: 1M  socket: 2  numanode: 2  core: 36  thread: 36
Utilized   Resources: PROCS: 36  SWAP: 10G
Dedicated  Resources: PROCS: 36  MEM: 72G
Attributes:         kernel=3-10-0-1160-42-2-el7-x86_64,puppethpc=20230110-09-49-143-15git52e6367,puppethpccode=20230111-6124-git6a9404cf,sudo=1-8-23-10-el7_9-2
  MTBF(longterm):   INFINITY  MTBF(24h):   INFINITY
Opsys:      linux     Arch:      ---
Speed:      1.00      CPULoad:  35.960
Partition:  pbs  Rack/Slot:  26/14  NodeIndex:  26
Features:   singleisland,beeond,cascadelake,type_cascadelake,r26,r26i13,genius,mem192
NodeType:   cascadelake
Classes:    [q1h][q24h][q72h][q7d][q21d][qsuperdome]
RM[pbs]*:   TYPE=PBS  STATE=Drained
NodeAccessPolicy: SHARED
EffNodeAccessPolicy: SINGLEJOB
Successive Job Failures: 1

Total Time:   INFINITY  Up:   INFINITY (98.77%)  Active:    919days (76.57%)

Reservations:
  51366877x1  Job:Running  -2:21:03:23 -> 2:02:56:37 (5:00:00:00)
Jobs:        51366877

source/leuven/genius_quick_start.rst
moravveji: On AMD nodes, we have still the singleuser policy:
```
$ checknode r28i39n01 | grep -i policy
EffNodeAccessPolicy: SINGLEUSER
```
and as Mag correctly demonstrated, only cascadelake nodes have shared policy.
Jan suggests to keep the current MOAB setting as it is, and only change everything consistently when we move to Slurm.
I will revise the text accordingly.

moravveji: and b.t.w. where do you see the `MINRESOURCE` phrase? I have not seen this before.
Can you plz give an example how to retrieve this?

MaximeVdB: Please keep in mind that "node allocation policy" and "node access policy"
are two different (though related) things ;).

The node allocation policy is the more simple of the two, and you can e.g.
find documentation about it here (including the `MINRESOURCE` mode):
https://docs.adaptivecomputing.com/maui/5.2nodeallocation.php

The node access policy is a more complex one. Strictly speaking I would call this
a property of a job  (not of a node or partition, although node- or partition-
specific settings are also influential here). This is because (at least on Genius)
it can also depend on the job resources (i.e. how many cores and how much memory)
what the job's node access policy will be (i.e. "shared" or "singlejob").
As e.g. Mag pointed out, there are essentially two different parts of Genius in
this regard: 
* one part where the jobs' node access policy is always "singlejob"
(the skylake "batch" and "bigmem" nodes, as well as the AMD nodes)
* another part where the same policy is applied, unless the amount of
cores and memory is within certain limits (then the "shared" node access
policy is given). This applies to all cascadelake nodes and all GPU nodes,
as well as Superdome.

(disclaimer: not sure if even this is a complete explanation, it could be even
more subtle)

MaximeVdB: And when updating this quickstart page, keep in mind that there also this other page,
(perhaps you can refer to it, or put some info there instead of in the quickstart page):

https://docs.vscentrum.be/en/latest/jobs/specifying_resources.html#enforcing-the-node-specification

source/leuven/genius_quick_start.rst
```diff
@@ -71,9 +71,18 @@ walltimes of 3 days or less.
...
+This means the CPU and memory resources per nodes are exploited as much as possible by packing more and more jobs into a single node.
+The ``SHARED`` node policy leaves room for smaller jobs to hit the queue and start earlier than scheduled initially.
```
stevenvdb: What do you mean by 'hitting the queue' in this context?

source/leuven/genius_quick_start.rst
stevenvdb: To be honest, I am more confused than before reading this PR (mostly because I thought the situation was much simpler). If I now understand things correctly, submitting a job with:
```
qsub -l nodes=1:ppn=4:skylake
```
will always lead to a situation where no other jobs can run on the node where my job runs? In that case, isn't it better to recommend users to request all cores, so they automatically get all the memory on the node and have more options with respect to pinning/binding? In both cases they will be _blocking_ an entire node.

Of course the different behavior of skylake and cascadelake complicates things a lot, so for simplicity towards the user, Ehsan's proposed text is fine for me. Also note that this will become irrelevant in a few months, when we switch to Slurm on genius.

moravveji: I tired to 
- avoid using the ``SHARED`` and ``SINGLEUSER`` jargons, but similar to the other documentation page, I mention nodes access policy being 'shared' or 'singleuser' based on what users can retrieve from `checkjob` (although that is tricky)
- made a simple distinction between skylake vs cascadelake nodes regarding the node access policy, with basic `qsub` examples
- I will include an additional sentence how to enforce getting skylake nodes using the architecture as a feature (in response to Steven's comment below).

source/leuven/genius_quick_start.rst
```diff
@@ -71,9 +71,29 @@ walltimes of 3 days or less.
 
 Submit to a compute node
 ~~~~~~~~~~~~~~~~~~~~~~~~
-To submit to a compute node it all boils down to specifying the required number of nodes and cores. As the nodes have a single user policy we recommend to always request all available cores per node (36 cores in this case). For example to request 2 nodes with each 36 cores you can submit like this::
+Submitting a compute job boils down to specifying the required number of nodes, cores-per-node, memory and walltime.
+The node access policy on Skylake and AMD nodes is 'singleuser'.
+This means that once a job lands on a Skylake node, no job from other users can land on the same node(s).
```
MaximeVdB, line 76:
```suggestion
This means that once a job lands on a Skylake or AMD node, no job from other users can land on the same node(s).
```

source/leuven/genius_quick_start.rst
MaximeVdB: It's definitely looking better now! But considering the Tier-2 audience, I think it would
be better to present this in a more gradual way, by first giving the basic info and only
give the more advanced info in a second step.

The contents of the basic part would include:
* an example for requesting two full nodes (the one that's already in there)
* recommending that "request only as much resources as needed by their applications"
  and there include e.g. your 8-core example
* to **not** yet mention the ":skylake"/"-l feature=skylake" (or cascadelake) stuff
  (and also not the node access policy etc). Just mentioning that the above example jobs 
  may end up running on either a Skylake or a Cascadelake node. 

The contents of the more advanced part would include:
* mentioning that users can control the node type with those ":skylake"/"-l feature=skylake"
  (and cascadelake) flags
* explaining the difference in the Skylake and Cascadelake compute nodes when
  it comes to node access policy. Here you can then mention that for small (few-core) jobs,
  it is best to either not select a specific nodetype feature or to select the Cascadelake feature.
* mentioning that there are also AMD compute nodes, but that those are described in a separate
  section below. If you want you can then also describe the node access policy in that AMD
  section.

source/leuven/genius_quick_start.rst
```diff
@@ -71,9 +71,19 @@ walltimes of 3 days or less.
...
 
    $ qsub -l nodes=2:ppn=36  -l walltime=2:00:00  -A myproject  myjobscript.pbs
+
+However, you may request a portion of resources on a node.
+For instance, to test a multi-threaded application which performs optimally using 4 cores, you may submit your job like this:
+
+   $ qsub -l nodes=1:ppn=4  -l walltime=2:00:00  -A myproject  myjobscript.pbs
+   
+In the two above examples, the jobs may start on Skylake or Cascadelake nodes.
+Please bear in mind no to exceed the maximum allowed resources on compute nodes for the targeted partition.
```
MaximeVdB, line 85:
```suggestion
Please bear in mind to not exceed the maximum allowed resources on compute nodes for the targeted partition.
```

source/leuven/genius_quick_start.rst
```diff
@@ -71,9 +71,19 @@ walltimes of 3 days or less.
...
+E.g. you can use maximum 36 cores per node ``ppn=36``.
```
MaximeVdB, line 86:
```suggestion
E.g. you can request at most 36 cores per node (``ppn=36``).
```
(bit easier to read)

source/leuven/genius_quick_start.rst
```diff
@@ -94,8 +104,24 @@ To specifically request V100 GPUs (which are on nodes with CascadeLake architect
   
 For the V100 type of GPU, it is required that you request 4 cores for each GPU. Also notice that these nodes offer much larger memory bank.
 
-Advanced usage
-^^^^^^^^^^^^^^
+Advanced node usage
+^^^^^^^^^^^^^^^^^^^
+The node access policy on Skylake and AMD nodes is `SINGLEUSER`.
```
MaximeVdB, line 109:
```suggestion
The node access policy on Skylake nodes is `SINGLEUSER`.
```

(I still think it would be confusing to mention the AMD nodes in this section
about the "batch" partition. In case you want to tell something about the  AMD
nodes, I think it's best to put that in the "Submit to an AMD node" section.)

source/leuven/genius_quick_start.rst
```diff
@@ -94,8 +104,24 @@ To specifically request V100 GPUs (which are on nodes with CascadeLake architect
   
 For the V100 type of GPU, it is required that you request 4 cores for each GPU. Also notice that these nodes offer much larger memory bank.
 
-Advanced usage
-^^^^^^^^^^^^^^
+Advanced node usage
```
MaximeVdB, line 107: It's a bit weird to see this "Advanced usage" (sub?)section in here,
I would have expected it to see it in the "Submit to a compute node"
section (after the 'basic part' in that section). Wouldn't it be better
like that?

I.e.

- Submit to a compute node
  (basic info)
  - Advanced usage
    (advanced info)
- Submit to a GPU node
  (basic info)
  - Advanced usage
    (advanced info)

MaximeVdB (changes requested):

source/leuven/genius_quick_start.rst
moravveji: The last commit should have taken care of all that.

source/leuven/genius_quick_start.rst
```diff
@@ -71,10 +71,35 @@ walltimes of 3 days or less.
...
 Submit to a compute node
 ~~~~~~~~~~~~~~~~~~~~~~~~
-To submit to a compute node it all boils down to specifying the required number of nodes and cores. As the nodes have a single user policy we recommend to always request all available cores per node (36 cores in this case). For example to request 2 nodes with each 36 cores you can submit like this::
+Submitting a compute job boils down to specifying the required number of nodes, cores-per-node, memory and walltime.
+You may e.g. request two full nodes like this:
 
    $ qsub -l nodes=2:ppn=36  -l walltime=2:00:00  -A myproject  myjobscript.pbs
-  
+
+However, you may request a portion of resources on a node.
```
MaximeVdB, line 79:
```suggestion
You may also request only a part of the resources on a node.
```
(makes it follow more logically from the preceding part)

source/leuven/genius_quick_start.rst
```diff
@@ -71,10 +71,35 @@ walltimes of 3 days or less.
 
 Submit to a compute node
 ~~~~~~~~~~~~~~~~~~~~~~~~
-To submit to a compute node it all boils down to specifying the required number of nodes and cores. As the nodes have a single user policy we recommend to always request all available cores per node (36 cores in this case). For example to request 2 nodes with each 36 cores you can submit like this::
+Submitting a compute job boils down to specifying the required number of nodes, cores-per-node, memory and walltime.
+You may e.g. request two full nodes like this:
```
stevenvdb, line 75: End the line with two colons in order to render what is below as code.

source/leuven/genius_quick_start.rst
```diff
@@ -71,10 +71,35 @@ walltimes of 3 days or less.
...
 
    $ qsub -l nodes=2:ppn=36  -l walltime=2:00:00  -A myproject  myjobscript.pbs
-  
+
+However, you may request a portion of resources on a node.
+For instance, to test a multi-threaded application which performs optimally using 4 cores, you may submit your job like this:
```
stevenvdb, line 80: End the line with two colons in order to render what is below as code.

source/leuven/genius_quick_start.rst
```diff
@@ -71,10 +71,35 @@ walltimes of 3 days or less.
...
+   
+In the two above examples, the jobs may start on Skylake or Cascadelake nodes.
+Please bear in mind to not exceed the maximum allowed resources on compute nodes for the targeted partition.
+E.g. you can request at most 36 cores per node (``ppn=36``).
+
+Advanced node usage
+^^^^^^^^^^^^^^^^^^^
+The node access policy on Skylake nodes is `SINGLEUSER`.
+This means that once a job lands on a Skylake node(s), no job from other users can land on the same node(s).
+If you insist on using a full node (to exclude jobs from other users), you may enforce getting a Skylake node in one of the following ways
```
stevenvdb, line 92: End the line with two colons in order to render what is below as code.

source/leuven/genius_quick_start.rst
```diff
@@ -71,10 +71,35 @@ walltimes of 3 days or less.
...
+
+   $ qsub -l nodes=1:ppn=8:skylake -l walltime=30:00 -A myproject myjobscript.pbs             # or
+   $ qsub -l nodes=1:ppn=8 -l feature=skylake -l walltime=30:00 -A myproject myjobscript.pbs
+
+The node access policy on Cascadelake nodes is `SHARED`.
+This means the CPU and memory resources per nodes are exploited as much as possible by packing more and more jobs into a single node.
+Similarly, you may enforce getting a Cascadelake node by specifying either `nodes=1:ppn=8:cascadelake` or `-l feature=cascadelake`
```
stevenvdb, line 99: I think the formatting here is nicer when you but double back quotes around the resource specifications, so \`\`nodes=1:ppn=8:cascadelake\`\` and  \`\`-l feature=cascadelake\`\`

stevenvdb (changes requested):
Content-wise this is now fine for me, but I have added some comments to improve the formatting.

source/leuven/genius_quick_start.rst
```diff
@@ -71,10 +71,35 @@ walltimes of 3 days or less.
...
+when only needing 8 cores.
+The `SHARED` node access policy leaves room for smaller jobs to start executing earlier than initially scheduled.
+Therefore, users are adviced to request only as much resources as needed by their applications.
```
MaximeVdB, line 102: As mentioned earlier, the node access policy is not the property of a node, since it is also a function of the amount of resources that a job requests. And one has to be careful to not call something a "node access policy" when actually referring to something that would rather be called the "node allocation policy". So I suggest the following:

```suggestion
E.g. you can request at most 36 cores per node (``ppn=36``). In general, we advise you to only request as much resources as needed by your application.

Advanced node usage
^^^^^^^^^^^^^^^^^^^
In certain cases (such as performance tests) you may want to be sure that your job runs on a specific type of node (i.e. only Skylake nodes or only Cascadelake nodes). You can do this by selecting a node feature via e.g. `-l nodes=1:ppn:8:skylake` or `-l feature=skylake` (same for `cascadelake`).

When doing so, you should take into account that all jobs on the Skylake nodes are subjected to the `SINGLEUSER` node access policy. This means that once a Skylake node is allocated to a job, no job from other users can land on this node (even if the original job only requested a small part of the node's resources). This is different on the Cascadelake nodes, where small jobs (less than 18 cores and with default memory requirements) are given the `SHARED` node access policy instead. This allows multiple small jobs from different users to run on the same node.
```

source/leuven/genius_quick_start.rst
moravveji: Thanks @stevenvdb for the formatting suggestions. I fixed them all.

source/leuven/genius_quick_start.rst
```diff
@@ -125,10 +150,10 @@ The big memory nodes are also located in a separate partition. In case of the bi
 
 Submit to an AMD node
 ~~~~~~~~~~~~~~~~~~~~~
-The AMD nodes are in their own partition.  Besides specifying the partition,
-it is also important to specify the memory per process (``pmem``) since
-the AMD nodes have 256 GB of RAM, which implies that the default value is
-too high, and your job will never run.
+The AMD nodes are in their own partition, with `SINGLEUSER` node access policy.
```
MaximeVdB, line 153: In a similar spirit as the suggestion above:

```suggestion
All jobs on AMD nodes are given the `SINGLEUSER` node access policy
(see above "Advanced node usage" paragraph for more information).

MaximeVdB (changes requested):

MaximeVdB (approved):

stevenvdb (approved):

PR #304, opened 2023-03-13
Update environment export behavior on wICE (plus Worker and Conda tips)

MaximeVdB:
This updates the `Environment propagation` paragraph in the wICE advanced guide, now that the CLI filter is in place.

### DIFF
--- a/source/leuven/wice_advanced_guide.rst
+++ b/source/leuven/wice_advanced_guide.rst
@@ -63,9 +63,9 @@ general information on software development on the VSC, have a look at this
 Worker
 ------
 
-The :ref:`worker framework <worker framework>`, which allows to conveniently
+The :ref:`Worker framework <worker framework>`, which allows to conveniently
 parameterize simulations, is available on wICE. An attention point is that
-if you want to lauch worker jobs from the Genius login nodes, you will need to
+if you want to lauch Worker jobs from the Genius login nodes, you will need to
 use a specific module:
 
 ::
@@ -77,7 +77,15 @@ If instead you want to launch Worker jobs from an interactive job running on
 wICE, you can use the ``worker/1.6.12-foss-2021a`` module (but do make sure
 this is the version installed in a subdirectory of ``/apps/leuven/icelake``).
 
-Currently ``-master`` flag does not work neither does ``wresume`` and all the resources have to be specified inside the `slurm script for worker <https://github.com/gjbex/worker/tree/development_slurm/examples/>`__ and not from the command line.
+Also note that the Worker support for Slurm is not yet complete. Both
+the ``-master`` option for ``wsub`` and the ``wresume`` tool currently
+only work for PBS/Torque and hence should not be used in the case of Slurm.
+
+All the resources furthermore need to be specified inside the Slurm script
+used as input for Worker (passing resources via the command line is not
+supported). Various examples can be found in a `development branch
+<https://github.com/gjbex/worker/tree/development_slurm/examples/>`__.
+
 
 .. _wice_monitoring:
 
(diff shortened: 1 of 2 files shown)


source/leuven/wice_advanced_guide.rst
```diff
@@ -108,6 +108,26 @@ For getting a compact overview of the current state of the cluster, execute
...
+Slurm jobs now start in a clean environment which corresponds to your login
+environment, i.e. with only those additional variables that you defined in your
+``~/.bashrc`` file. Environment variables that happen to be set in the session
+from which you submit the job are no longer propagated to the job.
+
+If needed you can modify this default behaviour with the
+`--export option <https://slurm.schedmd.com/sbatch.html#OPT_export>`__.
+When doing so, keep in mind that you will need to include the default minimal
+environment as well. To e.g. pass an additional environment variable ``FOO``
+with value ``bar``, use ``--export=HOME,USER,TERM,PATH=/bin:/sbin,FOO=bar``.
```
stevenvdb, line 125: Something weird is going on and I am not sure my findings are correct, but to me it seems that if you do not specify the `--export` option, Slurm defaults to `--export=HOME,USER,TERM,PATH=/bin:/sbin`. In case your job script has the `#!/usr/bin/bash` shebang, updates to `$PATH` in your `~/.bashrc` are ignored. On the other hand, if you have  `#!/usr/bin/bash -l` instead, the updates to `$PATH` are included.

Perhaps a good question is whether setting `PATH=/bin:/sbin` by default in a job is sensible. So instead of adapting the documentation, it might be better to first figure this out.

source/leuven/wice_advanced_guide.rst
```diff
@@ -108,6 +108,26 @@ For getting a compact overview of the current state of the cluster, execute
 ``slurmtop`` on any KU Leuven Tier-2 node. Use ``slurmtop --help`` to get to
 know the functionality.
 
+.. _wice_environment_propagation:
+
+Environment propagation
+-----------------------
+
+Slurm jobs now start in a clean environment which corresponds to your login
```
stevenvdb, line 116: This is nitpicking, but a variable like `$VSC_ARCH_LOCAL` will not have the same value as in the login environment. I am not sure if adding a remark about this would clarify things or just make them more complicated, so you can decide if you want to add it.

stevenvdb (changes requested):

source/leuven/wice_advanced_guide.rst
MaximeVdB: For the time being, for sure the easiest is to inform the users that it's best to use  `#!/usr/bin/bash -l` instead of  `#!/usr/bin/bash`. Currently the use of `bash -l` is explicitly mentioned for interactive jobs (https://docs.vscentrum.be/en/latest/leuven/wice_quick_start.html#submit-to-the-interactive-partition). I will add a commit here to make clear how the shebang is supposed to look like for batch jobs.

source/leuven/wice_advanced_guide.rst
MaximeVdB: That's only if you define "login environment" as "environment you would get upon logging in to a login node" ;). Here I meant "login environment" in the context of a job, so as "login environment when logging in to an allocated compute node". For now I think this is an OK description that does not overcomplicate it.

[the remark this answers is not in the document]
MaximeVdB: (done in 3f99a20b05c9ab0646cf6b2edbf1722d47251177)

source/leuven/wice_advanced_guide.rst
MaximeVdB: Done in 7baf31678d6a191c2342120ad373d6ba3aa3b23a

stevenvdb (approved):

mselwa (approved):
Sorry for adding 2 line about worker and a conda section. Please check if you agree with that.

PR #305, opened 2023-03-18
add github action to test sphinx build

smoors:


.github/workflows/main.yml
```diff
@@ -0,0 +1,16 @@
...
+
+jobs:
+  build_docs_job:
+    runs-on: ubuntu-22.04
+    name: build sphinx docs
+    steps:
+      - uses: actions/checkout@v3
+      - name: install sphinx
+        run: |
+          python3 -m pip install sphinx==4.3.0
```
lexming, line 12: I would use the `requirements.txt` file in here as well, then we can have a single source for the build environment.

lexming (approved):
The action seems fine. Although I know for a fact that the current documentation does not build cleanly, so we should not block PRs due to failed tests (yet).

smoors:
> The action seems fine. Although I know for a fact that the current documentation does not build cleanly, so we should not block PRs due to failed tests (yet).

it indeed does not build cleanly (currently ~370 warnings), but the build current succeeds (no nonzero error code).
see my account, where i already enabled it:
https://github.com/smoors/VscDocumentation/actions/runs/4466854954/jobs/7845568013

so i don't see why not to block PRs due to failed test.

lexming:
I was expecting the warnings to fail the test. Then it's good, maybe after the pydata update we can also make warnings fatal.

stevenvdb (approved):
Seems ok to me. The only thing I am wondering is whether the build here is done in exactly the same way as it is done when the pages are put into production, but I am not sure where that information can be found.

.github/workflows/main.yml
```diff
@@ -0,0 +1,16 @@
+name: test_sphinx_build
+on: [push, pull_request]
+
+jobs:
+  build_docs_job:
+    runs-on: ubuntu-22.04
```
moravveji, line 6: This PR looks good to me.
Just out of curiosity, are we freezing the build only on this specific version of ubuntu, or `ubuntu-latest` could be an alternative, for a longer run into the future?

moravveji (approved):
LGTM

.github/workflows/main.yml
smoors: it's better to keep this fixed, so if the build fails we know it's not due to the ubuntu version used.
22.04 is a long time release, so we're good for a few more years (can always be updated later of course).

PR #309, opened 2023-04-07
Update Superdome info w.r.t. Slurm migration

MaximeVdB:


source/leuven/superdome_quick_start.rst
```diff
@@ -1,77 +1,42 @@
...
-The :ref:`Superdome <Superdome hardware>` is a shared memory machine, i.e.,
-it can be used to run multi-threaded application that require large amounts
-of RAM.
-
-
-How to connect to Superdome?
-----------------------------
-
-Superdome does not have a dedicated login node, so in order to work with the
-system users can connect to the same login node used for Genius.
```
moravveji, line 13: Just for the sake of completeness, please provide a link to the Genius login node reference page with the anchor `tier2_login_nodes`.

source/leuven/superdome_quick_start.rst
```diff
@@ -1,77 +1,42 @@
...
+the ``superdome`` or ``superdome_long`` Slurm partition, depending on whether
+your job needs less than 3 days or less than 7 days.
 
-- specify the partition ``superdome``,
-- specify the queue ``qsuperdome``, and
-- use ``-L`` for specifying resources.
+The CPU resources on Superdome are furthermore allocated on the basis of entire
+sockets and not on the basis of individual cores as on other partitions.
+Since the 112 cores in Superdome are grouped in 8 sockets (see the
+:ref:`Superdome hardware page<Superdome hardware>`), this means that all
```
moravveji, line 20: This link is not correctly rendering.

source/leuven/superdome_quick_start.rst
```diff
@@ -1,77 +1,42 @@
...
-For instance, if you would have a sequential program that required 500 GB of
-RAM, you would use ``-L tasks=1:lprocs=1:place=numanode=1:memory=500gb``.
+.. note::
 
-However, keep in mind that if the memory you require exceeds that of a
-NUMA-node, you would have to request multiple nodes to accommodate for
-the total amount of memory.  For example, if a sequential program would
-require 1 TB of RAM, you would specify:
-``-L tasks=1:lprocs=1:place=numanode=2:memory=1tb``.
+   The maximum amount of memory per core for Superdome is 53.5 GB, which is also
```
moravveji, line 39: Well, `scontrol` tells me that `MinMemoryCPU=53500M`. When "properly" converting MB to GB (using 1024 factor), we arrive "safely" at 52GB. But, I do not know which conversion factor is assumed under the hood by Slurm between MB and GB.

source/leuven/superdome_quick_start.rst
```diff
@@ -1,77 +1,42 @@
...
+   the default value. Your jobs will hence by default receive 749 GB of RAM per
```
moravveji, line 40: depending on the above remark, this 749GB also might need to be revisited.

source/leuven/tier2_hardware/superdome_hardware.rst
```diff
@@ -1,19 +1,20 @@
 Superdome hardware
 ==================
-Superdome is a shared memory machine and consists of several nodes.
-
-
-.. include:: tier2_login_nodes.rst
-
+Superdome is a single node with a large amount of memory (6 TB)
```
moravveji, line 3: I would have written it like:
"Superdome is a single-node machine equipped with 6 TB of memory, and is part of the Genius cluster."

source/leuven/tier2_hardware/superdome_hardware.rst
```diff
@@ -1,19 +1,20 @@
...
+Superdome consists of
+
+- 8 sockets with
+
+    - Xeon Gold 6132 CPU\@2.6 GHz (Skylake), 14 cores each
+    - 750 GB RAM each
 
-    - Xeon Gold 6132 CPUs\@2.6 GHz (Skylake), 14 cores each
-    - 750 GB RAM
-    - partition ``superdome``
```
moravveji, line 15: - partition ``superdome`` and ``superdome_long``

moravveji (changes requested):
Thanks for taking care of these two pages.

source/leuven/superdome_quick_start.rst
MaximeVdB: Could you double check? When I render it and view the page the link looks fine and points to expected destination.

source/leuven/superdome_quick_start.rst
MaximeVdB: Yes, I see that I somehow moved that part to the Superdome hardware page instead of here, I'll put it back (but just a reference to the Genius login infopage, no need to repeat the whole thing).

source/leuven/superdome_quick_start.rst
MaximeVdB: https://slurm.schedmd.com/slurm.conf.html talks about megabytes (MB) and those are a decimal system. So 53500 MB = 53.5 GB. The factor of 1024 would enter if you'd talk about MiB and GiB instead. So to me this looks good. But I also had to double check ;)

source/leuven/superdome_quick_start.rst
MaximeVdB: done (34d94295f9fc4bbc8a30f23b80fdc2e510192267)

moravveji: exactly. that will do

source/leuven/tier2_hardware/superdome_hardware.rst
MaximeVdB: Indeed better like that (done in 34d94295f9fc4bbc8a30f23b80fdc2e510192267)

source/leuven/tier2_hardware/superdome_hardware.rst
MaximeVdB: I don't think scheduler partition info should be copied to a hardware info page.
There is already a reference to the Superdome quickstart page where the Slurm partition info is provided.

source/leuven/superdome_quick_start.rst
moravveji: yes, it renders correctly when I fully make the docs locally on my laptop.

stevenvdb: For me this is also ok.

source/leuven/tier2_hardware/superdome_hardware.rst
moravveji: Well, one of the two partitions is already mentioned here, so, I do not see anything against mentioning the 2nd valid partition name. It is very relevant to this page too.

Unfortunately, in the Genius hardware page, we are only mentioning the 3-day partition names, and the 7-day partitions are only explained in the quick start page. The same for wICE. 
To me this is not completely OK, but this is a matter of taste.

What if a user comes up with a question like: "what are the possible/available partitions on Genius?". then on the VSCdocs, he does not find it all in the Genius hardware page; he only finds a partial answer. And without knowing about the other `_long` partitions, the user will stop searching, so, he might not figure it out.

I propose that we rename and slightly reformulate (of course in a separate PR) the Genius|wICE hardware pages, and call them like "Genius|wICE Hardware and Configuration" pages. Then, we can safely mention e.g. `gpu_p100_long` partition in that page, too.

But for now, the fact that `superdome_long` is not mentioned in the hardware page also complies with other Genius hardware page.

So, this PR is good to GO.

moravveji (approved):

source/leuven/superdome_quick_start.rst
```diff
@@ -1,77 +1,44 @@
...
-   The number of logical processors is essentially the number of cores
-   you want to use for your job.
-``place``
-   The place determines where the logical processors are executed, and
-   is ``numanode`` for Superdome.  Each ``numanode`` has 14 cores and 750 GB of
-   RAM, so if you want to use, e.g., 28 cores, you would specify
-   ``lprocs=28:place=numanode=2``.
+  $ sbatch -M genius -p superdome -A myaccount --ntasks=1 myscript.slurm
 
+The following job will get two sockets (and so 28 cores in total)::
```
stevenvdb, line 30: By default, each task will be launched on a separate socket. The following job will therefore get two...

stevenvdb (approved):
The documentation seems ok.

I tried to do something stupid and submit with `--ntasks-per-socket=9` and `--ntasks=19`. Slurm actually handled this nicely by providing me 3 sockets and launching only 7 tasks on the two first sockets and 5 tasks on the last one.

Did you by any chance test worker on superdome?

source/leuven/superdome_quick_start.rst
MaximeVdB: Yes, would be good to mention that already there (done in 592b31402e100be571bcb942ed54d1424c33d1be)

MaximeVdB:
> Did you by any chance test worker on superdome?

No, not yet. If Superdome-specific-Worker-tips turn out to be needed, these can be added in a next PR. With the current one at least the basics are covered :).

PR #313, opened 2023-04-14
update hydra hardware

smoors:


source/brussels/tier2_hardware/hydra_hardware.rst
```diff
@@ -20,18 +20,19 @@ ivybridge_mpi    16      2x 10-core INTEL E5-2680v2 (Ivybridge)      256 GB  900
...
-===============  ======  ===============================  ==========  =======================================  ==========  ==========  =======
+==========================  ======  ===============================  ==========  =======================================  ==========  ==========  =======
+Slurm partition             nodes   GPUs per node                    GPU memory  processors per node                      CPU memory  local disk  network
+==========================  ======  ===============================  ==========  =======================================  ==========  ==========  =======
+pascal_gpu                  4       2x Nvidia Tesla P100 (Pascal)    16 GB       2x 12-core INTEL E5-2650v4 (Broadwell)   256 GB      2 TB HDD    10 Gbps
+ampere_gpu                  6       2x Nvidia A100 (Ampere)          40 GB       2x 16-core AMD EPYC 7282 (Zen2)          256 GB      2 TB HDD    EDR-IB
+ampere_gpu (big_local_ssd)  4       2x Nvidia A100 (Ampere)          40 GB       2x 16-core AMD EPYC 7282 (Zen2)          256 GB      5.9 TB SSD  EDR-IB
```
lexming, line 34: Better move the *big_local_ssd* bit in its own column called *Features* to avoid users using that as a partition name.

lexming (changes requested):

source/brussels/tier2_hardware/hydra_hardware.rst
smoors: we should split the table in 2 parts as it's too much info to fit nicely, but that will be for another day.

hajgato (approved):

PR #314, opened 2023-04-14
Restoring the old hardware pages

moravveji:
I have restored the `.rst` pages about Superdome, ThinKing and Breniac from previous commits. The restored files are copied into  `sources/leuven/old_hardware` for possible future reference.
It would be good to keep doing this later on as well, when in the future our older machines are decommissioned.


stevenvdb:
Overall I can live with keeping old hardware pages, but I do have a few remarks:
- Since these pages are still accessible through the search function, I would clearly mark them as outdated. A note on top of the page specifying this concerns hardware that is no longer in use would suffice.
- You have added Superdome, which is still alive and kicking and is already linked to on our Tier-2 Hardware page
- Some references in these pages are no longer present (for instance `:ref:'options <submit to Breniac skylake node>'`) which gives `WARNING: undefined label` messages when building the pages. At this moment this is not dramatic, but if we in the future automate checks for this it can become cumbersome.

stevenvdb:
Now `breniac_hardware.rst` includes `breniac_login_nodes.rst` and both have the warning about being obsolete. So that message is now shown twice on the Breniac hardware page. Since these pages are only retained for reference, I don't mind too much.

stevenvdb (approved):

gjbex (approved):

PR #315, opened 2023-04-25
Add Slurm equivalent for -l nodes=...:ppn=...

MaximeVdB:
Quite some KU Leuven users who were used to PBS have found the "Converting PBS/Torque options to Slurm" page, which is good. But then they focus on the `-L tasks=<X>:lprocs=<Y>` example, which is typically not the right one for converting e.g. their old Genius PBS jobs scripts to Slurm. So this PR adds a `-l nodes=...:ppn=...` example in the hope that KU Leuven users will pick it up.

@RVerschoren I hope this small addition is OK for UAntwerpen as well. In the longer term (e.g. in the future layout for VscDocs) we'll presumably have a better distinction between e.g. common PBS-to-Slurm info and UA- and KU Leuven-specific info.


WouterVanAssche (approved):

MaximeVdB:
@RVerschoren @backelj  so is this OK for UA?

RVerschoren:
We will likely still recommend the `--ntasks` form, but the added info can't really hurt in this case.

RVerschoren (approved):
lgtm

PR #322, opened 2023-05-24
add info on inode quota

smoors:


source/access/how_to_request_more_quota.rst
```diff
@@ -11,9 +11,32 @@ option depends on data storage policies of the site managing your VSC account,
...
+If you need more :ref:`inode quota <quota>`, you should first try to
+reduce the number of files as much as possible:
+
+- Clean up the files you no longer need.
+- Pack the remainder in a tarball, e.g. ``tar zcf myfiles.tgz file1 <...>
+  fileN``. If you need help on ``tar``, please consult the man page, *i.e.*, do:
+  ``man tar``.
+- If tarballs are not a suitable option for you, consider agregating your data
+  into large ``CSV``, ``HDF5``, or ``netCDF`` files, or using a database such as
+  ``SQLite``, ``DuckDB``, ``MariaDB``, or ``PostgreSQL``.
```
stdweird, line 31: duckdb? never heard of it before this PR; not packaged in el8 and not in EB? 
why do you bring this up here?

source/access/how_to_request_more_quota.rst
```diff
@@ -11,9 +11,32 @@ option depends on data storage policies of the site managing your VSC account,
...
+===========================
+
+If you need more :ref:`inode quota <quota>`, you should first try to
+reduce the number of files as much as possible:
+
+- Clean up the files you no longer need.
+- Pack the remainder in a tarball, e.g. ``tar zcf myfiles.tgz file1 <...>
+  fileN``. If you need help on ``tar``, please consult the man page, *i.e.*, do:
+  ``man tar``.
+- If tarballs are not a suitable option for you, consider agregating your data
```
stdweird, line 29: so there is code that does a lot of small file readnig/writing, the user cannot change it to to wrap the file open in a tarball open en then a lookup to the file in the tarball, but somehow they can change to these suggestions? sounds a bit odd.

feel free to state that they should start with changing the code to use netdcdf or hdf5, since it's a much better format; and then worst case use tarballs (and worser case csv).
sqlite is also an option, but mariadb and postgres don't support direct db file access as far as i know.

source/access/how_to_request_more_quota.rst
```diff
@@ -11,9 +11,32 @@ option depends on data storage policies of the site managing your VSC account,
...
+.. _more quota inode:
+
+Requesting more inode quota
+===========================
+
+If you need more :ref:`inode quota <quota>`, you should first try to
+reduce the number of files as much as possible:
+
+- Clean up the files you no longer need.
+- Pack the remainder in a tarball, e.g. ``tar zcf myfiles.tgz file1 <...>
```
stdweird, line 26: most useful comment here would be to use a parallel zip instead of default, since it is by far the slowest part of this comment and most usefull when doing large cleanups

source/access/managing_disk_usage.rst
```diff
@@ -6,6 +6,22 @@ How much disk space am I using?
...
+Disk quota
+==========
+
+VSC file systems can have 2 types of quota set on them:
+
+* quota on the *disk space*
+* quota on the *number of inodes*
+
+The number of inodes determines the maximum number of files that can be present
+in the file system. The purpose of the inode quota is to maintain good I/O
```
stdweird, line 18: in the ugent tier2 setup and also on tier1 the total amount of inodes is finite like the regular storage. and this is the main reason why we have quota on them. it has nothing to do with performance.

source/access/how_to_request_more_quota.rst
```diff
@@ -11,9 +11,32 @@ option depends on data storage policies of the site managing your VSC account,
...
+- If tarballs are not a suitable option for you, consider agregating your data
+  into large ``CSV``, ``HDF5``, or ``netCDF`` files, or using a database such as
+  ``SQLite``, ``DuckDB``, ``MariaDB``, or ``PostgreSQL``.
+- Be careful when running jobs that write back to the shared storage and clean
+  up anything you do not need for post-processing or for verification, *i.e.*,
+  intermediate files.
+- Take care not to let your jobs write such small files on the shared storage
+  systems, but instead have them write to the local disks on the nodes where
+  your jobs are running and then pack them if they are still needed before
+  moving them to the shared storage.
```
stdweird, line 38: this paragraph is missing the part from the title

source/access/how_to_request_more_quota.rst
smoors: it's just a suggestion, just like the other options.
duckb is optimized for analytical workloads, which is probably what most researchers do.
it seems to be getting quite popular.
it's just a python package so probably not an issue to install.
no problem to remove it if you care so much, but why are you so against this?

stdweird: it's not just a python package, it's probably not available on the clusters at this moment, from the website it seems not optimal for HPC at all: why do you mention it? did you have some recent success from users trying it?

source/access/how_to_request_more_quota.rst
smoors: again, these are just suggestions.
i can imagine that tarballs are not a practical solution if the data need to be accessible for further analysis/processing
i did put the tarball option first because that is the simplest  (they don't have to change the code, just do an extra step afterwards). but no problem to change the order if you prefer.

for mariadb and postgres, i wasn't sure, i'll remove them

source/access/how_to_request_more_quota.rst
smoors: sorry, but i don't understand. what do you want exactly?

stdweird: The title is `Requesting more inode quota`, yet there is nothing in here about requesting more inode quota

source/access/managing_disk_usage.rst
smoors: i'm not a file system expert so correct me if i'm wrong, but i don't think this is correct.

if you wanted could increase the number of inodes by decreasing the bytes per inode?
so why don't you do that? because of performance? because random file access is expensive?

stdweird: in our gpfs and lustre config, the inodes are hosted on ssd or nvme, not on the spinning disks. so they are finite, and complicated to extend only this storage. (well, as complicated as the non-inode storage; so we have the same reasons for inode quota as for regular quota).

changing the inode size is not trivial, there is only small (512byte) and large (4kB); and the large inodes actually hold very small files; eg in gpfs on tier2 files with size under 4k-few 100 bytes of inode are actually fully stored in the inode. but that is not really the point

source/access/how_to_request_more_quota.rst
smoors: i don't have experience with it, nor have our users that i know if.
it's a popular sqlite alternative and is claimed to be more performant for analytical workloads as it is column-based instead of row-based like sqlite.
there is no one-size-fits-all in DBs. it may not be optimal for HPC, but neither is sqlite.

i'll remove it.

source/access/managing_disk_usage.rst
smoors: thank you for that info.

even so, you could have chosen a larger inode storage and small sizes from the start, so it's a design choice, and the reason for that choice is performance?

stdweird: performance has nothing to do with it. it's the same reason the regular quota exists: the resource is finite. ofcourse  it also helps to keep people from doing stupid things that might impact the whole system. 
but if people need insane amounts of small files, they can get them. but not by default, they will need to motivate why this is needed.

smoors: it's not just helping. the quota are an important safeguard to limit the amount of random IO, which indeed impacts the whole system. 
of course this is not enough, users can still do stupid things, but that's exactly why i wrote this, to make users aware of this limitation.

anyway, i give up. can you please provide an alternative text so i can get this merged?

source/access/how_to_request_more_quota.rst
smoors: something like this? `tar -I pigz cf myfiles.tgz file1 <...> `

stdweird: yes. we have pigz on all our nodes, so that is fine with me

source/access/how_to_request_more_quota.rst
stdweird: i would also not recomend sqlite tbh ;)

source/access/how_to_request_more_quota.rst
smoors: hm, it's not available in the leuven genius login

smoors: ok, they just installed it.
antwerpen also has it.

source/access/managing_disk_usage.rst
```diff
@@ -6,6 +6,28 @@ How much disk space am I using?
...
+==========
+
+VSC file systems can have 2 types of quota set on them:
+
+* quota on the *disk space*
+* quota on the *number of inodes*
+
+The number of inodes limits the maximum number of files that can be present in
+the file system (although there is no one-to-one relation between inodes and
+files).
```
stdweird, line 19: `files` -> `number of files`

source/access/managing_disk_usage.rst
```diff
@@ -6,6 +6,28 @@ How much disk space am I using?
...
+
+The inode and space quota allow site admins to control the amount of data and/or
+number of files users create, and to prevent system wide problems in case of
+unexpected problems, such as runaway jobs filling up the available space or
```
stdweird, line 23: hmm, maybe `problems` -> `behaviour`?

stdweird (approved):

stdweird:
@smoors 2 tiny suggestions, i already approved it

wpoely86 (approved):

lexming (approved):
LGTM

lexming:
@stdweird can you re-approve please? last commits invalidated your review

stdweird (approved):

hajgato (approved):

PR #327, opened 2023-06-01
Use appropriate memory per core value (Genius quickstart)

MaximeVdB:
Just something I noticed the other day and that would be good to fix.

(A "G" (or 'g') unit multiplies by 1024, so these two examples would not behave as expected (the first one would get rejected upon submission, the second one leads to additional cores being allocated to satisfy the memory requirement).)

### DIFF
--- a/source/leuven/genius_quick_start.rst
+++ b/source/leuven/genius_quick_start.rst
@@ -2,7 +2,6 @@
 
 Genius quick start guide
 ========================
-
 :ref:`Genius <Genius hardware>` is one of the two KU Leuven/UHasselt Tier-2 clusters,
 besides :ref:`wICE <wice hardware>`. 
 Given the architectural diversity of compute nodes on Genius, this cluster is suited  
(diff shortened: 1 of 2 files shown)


moravveji:
Clearly, this is the issue of memory unit conversion, trying to match the node-memory settings for different partitions. Fixing the docs is one approach, but adapting the Slurm partition configs is another approach, and personally, I prefer the latter.

While we used PBS/Torque, a single-core job e.g. on `bigmem` partition could request 20GB of memory and still start properly on the target node, with effectively one core assigned to the job. This means that requesting e.g. 20GB per core on `bigmem` partition leaves sufficient memory for the OS and other services, and can be fine. 

Slurm is based on `MB`, and all we need to do to fix both issues is to define the node memory specifications n multiples of 1024, rather than 1000. So, e.g. for the `bigmem` partition, it would be better `DefMemPerCPU=5120 MaxMemPerCPU=20480`.

That can let the jobs running as expected (like using a single core when asking for a single core), and gives less confusion to the users when submitting a job (specifying `20G` instead of `20000M`), and the VSC docs stays the same, and users stay using the nodes as they are taught how to, during the intro sessions.

moravveji (approved):
alright

MaximeVdB:
> alright

(just adding for the record that we'll look into adapting the Slurm configuration as suggested by Ehsan, but that we will first make the documentation consistent with the current setup)

stevenvdb (approved):
Looks good. Perhaps you can add a note saying that 20G is not equal to 20000M for Slurm, since some users might try the conversion themselves?

MaximeVdB:
> Looks good. Perhaps you can add a note saying that 20G is not equal to 20000M for Slurm, since some users might try the conversion themselves?

Yes, I initially thought of keeping that change for a more thorough revision (i.e. the one where we would also restructure w.r.t. the UAntwerpen Slurm pages etc.), but it's indeed probably best to already add it now. I did feel that adding this info in a proper way (done in 3f018915729a17ea15098f7ad1c8d87d562a38eb) first required some restructuring/polishing of the Genius quick start page (done in c304a4ee43c0ba6c11921b8224952a9c365939c6) -- I hope that's OK.

source/leuven/genius_quick_start.rst
```diff
@@ -15,61 +14,79 @@ For example, to log in to any of the login node using SSH::
...
+The `Slurm Workload Manager <https://slurm.schedmd.com>`_ is the scheduler,
+resource manager and credit accounting manager on Genius (and wICE).
 
-- :ref:`hardware specification <Genius hardware>`
-- Submitting jobs using :ref:`Slurm Workload Manager <Antwerp Slurm>` and the 
-  :ref:`Advanced topics <Antwerp advanced Slurm>`
-- :ref:`running jobs <running jobs>`
-- :ref:`obtaining credits <KU Leuven credits>` and 
-  :ref:`Slurm accounting <accounting_leuven>`
+In case you are not yet familiar Slurm and/or the Genius hardware, you can find
```
stevenvdb, line 28: familiar Slurm -> familiar with Slurm

source/leuven/genius_quick_start.rst
```diff
@@ -15,61 +14,79 @@ For example, to log in to any of the login node using SSH::
...
-   In general, we advise you to only request as much resources as needed by your application.
+   Please bear in mind to not exceed the maximum allowed resources on compute
+   nodes for the targeted partition. E.g. you can request at most 36 cores per
+   node (``--ntasks=36``). In general, we advise you to only request as many
+   resources as needed by your application.
+
+.. note::
+
+   If you do not provide a walltime for your job, then a default walltime will
+   be applied. This 1 hour for all partitions, except for the debug partitions
```
stevenvdb, line 72: This 1 -> This is 1

stevenvdb (approved):
Fine for me, only two small typos that could be fixed.

MaximeVdB:
> Fine for me, only two small typos that could be fixed.

Thanks, I've now fixed those in fb6320c3f5cb93b60cf41fa68279c7fe7e33b8a9

PR #339, opened 2023-08-30
slight wording correction is proposed

moravveji:
I propose a small wording correction in this page. 

But beside that, I have a more general comment.
The text is trying to convince us that we need "powerful supercomputers" to advance science. But, having such machines is not enough, without having efficient code and proficient users!
So, I propose to iron out this page again, and at least use "supercomputing" instead of "supercomputer" wherever it fits into the text.

Because this specific page is a bit philosophical (!), I am not gonna propose text, as it can trigger long discussions.


smoors (approved):
lgtm

lexming (approved):

PR #340, opened 2023-08-30
fix links to VSC website in Access section

lexming:
A few links to the VSC website in the access section are broken (thanks to @jantur3k for catching those :beers:)  . They use URLs from a named reference and lack a `_` to get the reference interpreted as such.

This is a weird error that results in a broken link but does not triger any warning or error at compile time. That's why it went through the cracks.

I'm also updating the styleguide to cover this case.

### DIFF
--- a/source/access/windows_client.rst
+++ b/source/access/windows_client.rst
@@ -67,8 +67,7 @@ system of the cluster.
 
 |Recommended| Use the X server included in :ref:`MobaXterm <access using mobaxterm>`.
 
-Alternatively, you can install an X server such as `Xming <Xming web site>`_ on
-Windows as well.
+Alternatively, you can install an X server such as `Xming`_ on Windows as well.
 
 .. toctree::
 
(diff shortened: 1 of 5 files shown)


source/access/windows_client.rst
```diff
@@ -67,7 +67,8 @@ system of the cluster.
 
 |Recommended| Use the X server included in :ref:`MobaXterm <access using mobaxterm>`.
 
-Alternatively, you can install an X server such as `Xming`_ on Windows as well.
+Alternatively, you can install an X server such as `Xming <Xming website_>`_ on
```
smoors, line 70:
```suggestion
Alternatively, you can install an X server such as `Xming <Xming website>`_ on
```

lexming: @smoors Read the description! :smile:  That underscore is precisely needed for links defined from a named reference :slightly_smiling_face:

smoors (approved):
lgtm

lexming:
@gjbex this one is ready, can you please merge it?

gjbex:
You don't need me to merge, you have the permissions to do it yourself.

lexming:
But we need 2 reviews :slightly_smiling_face:

PR #342, opened 2023-08-30
Add links to LinkedIn and VSC website to nav bar

### LINKED ISSUE #338: Title bar/header
The bar at the top of the page contains a number of useful (or not so much) icons:
* Does the one to Github add value?  We don't have much material there that is public/interesting
* Can we replace the Twitter icon with the X icon?
* Can we add a LinkedIn icon since that is our main social media channel?
* Can we add an icon for the VSC website?

lexming:
Fixes #338

Preview in https://vscdocumentationbeta.readthedocs.io

Adding the logo for X is quite hacky because it is not available in the vendored version of FontAwesome in the pydata theme. So I added it manually as an image an emulated the behaviour of the other icons with CSS.

Whenever we get a version of pydata with a recent enough FontAwesome, this can be replaced by the simple method using icons.

Regarding the link to the VSC website, I chose a home icon for it. I didn't find any better option in the free [FontAwesome collection](https://fontawesome.com/search?o=r&m=free) that would convey the message of a link to a parent website.


gjbex (approved):
Looks good to me, thanks.

smoors (approved):

smoors:
hm, now that i see the result, i think it would be better to let the VSC logo link to `vscentrum.be`, and let the home icon link to `docs.vscentrum.be`.

the home is usually the home page of the site you are currently on

would you agree with that change?

PR #343, opened 2023-08-31
Add custom logos for VSC docs and link to VSC website

lexming:
Follow-up to #342 

As suggested by @smoors, some logos used in the navbar are confusing:
* the top left logo is currently the default VSC logo, which does not convey the message that clicking it you return to frontpage of documentation
* the top right logo of a home does not convey the message that you will go to VSC website, and it conflicts with other home logos used in the documentations pages

This PR tries to fix this issue by using custom logos for those 2 buttons. I made those custom logos myself, so do not expect a lot of creative power here :wink: I hope that you like them and that their intention is are clear.

Preview in https://vscdocumentationbeta.readthedocs.io/ (note: you might need to reload the page resetting the cache with ctrl+shift+R)


lexming:
Update:
* using much better logo for VSC docs made by @smoors 
* adding the VSC favicon to the documentation website (thanks @gjbex)

gjbex (approved):
Excellent, looks much better.  I don't see the favicon in preview mode, but that is to be expected.

PR #348, opened 2023-09-11
Update genius_2_rocky.rst

moravveji:


stevenvdb (approved):
I think these proposed changes certainly clarify things.

I can propose some more things that in my opinion would improve this page:

* I think that the remark about the new modulepath hierarchy is a technical detail for the majority of users. Most users should just use the `cluster` modules from now on and not modify the `$MODULEPATH` variable. I would move the section called "Using cluster modules" into the section "Impact on centrally installed modules". The note about modulepath hierarchy can stay of course, but it should be made clear that this is more for advanced users.
* We should take care that the page is not already outdated when the migration happens. I would introduce a section called something like `Temporary migration remarks` and collect there information about for instance the pilot group. Shortly after the migration we can then remove it.
* During the initial stages of wICE, we recommended users to do `module --force purge`. If users still have that in their job scripts, they will not be able to load any modules since this will unload the cluster module that is now loaded automatically. I would add a specific warning to not use this statement anymore.

If you feel this does not belong in this PR, I am certainly willing to make a PR of my own after yours has been merged.

moravveji:
@stevenvdb Your suggestions are very relevant. Let's do them all in a separate PR once this is merged.

moravveji:
@stevenvdb : I bypassed Jan's review. You may open a 2nd PR for your proposed improvements.

PR #352, opened 2023-09-13
Revamp the Genius to Rocky 8 page

stevenvdb:
I have made some quite aggressive changes to the Genius to Rocky 8 page, I hope everyone is ok with that. My main objectives are:
* Provide a more extensive description of how the cluster module works
* Make sure most of this page remains valid _after_ the migration, since I think that the discussion on the cluster module should be permanent and not temporary.

I have a few remarks
* There is some overlap with other pages such as https://docs.vscentrum.be/software/software_stack.html or https://docs.vscentrum.be/gent/setting_up_the_environment_using_lmod_at_the_hpc_ugent_clusters.html# In the long term it would be good to merge it in there (although there can be some CUE issues)
* In a few other places, we recommend to execute `module use` to adapt the modulepath. This will no longer be needed after the migration and those pages should be adapted then.
* We now recommend users to have a separate conda installation with `-rocky` as suffix. On the wICE page however, we recommend the `-wice` suffix, so maybe we should try to discuss this at some point to make things more consistent (I would also prefer `-rocky8`, to avoid issues in the next OS update).

### DIFF
--- a/source/leuven/genius_2_rocky.rst
+++ b/source/leuven/genius_2_rocky.rst
@@ -61,7 +61,7 @@ If you only have a Conda environment working on Genius, it's best to create a ne
   
 In order to install miniconda in a new directory you can ::
 
-   bash Miniconda3-latest-Linux-x86_64.sh -b -p $VSC_DATA/miniconda3/rocky
+   bash Miniconda3-latest-Linux-x86_64.sh -b -p $VSC_DATA/miniconda3-rocky
    export PATH="${VSC_DATA}/miniconda3-rocky/bin:${PATH}
 
 You can then use this Conda environment after the migration. You can prepare this on the test nodes.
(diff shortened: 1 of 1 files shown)


source/leuven/genius_2_rocky.rst
```diff
@@ -1,121 +1,267 @@
...
-------------------------------------------
-In order to minimize the changes you need to make to your jobscripts, an appropriate module path (``$MODULEPATH``) will be set by default at the start of your job of the the migration of Genius to Rocky 8 OS. This new module path will now contain all toolchain versions, so all the modules, starting from 2018a on Genius and starting from 2021a on wICE. This is an important change! Previously the modulepath was set to a single toolchain, that would not change over time. You might have set a module path in your jobscripts to refer to newer toolchains. Module use/unuse should now only be needed in exceptional cases.
+.. _manually_modifying_modulepath:
 
-.. note::
+Manually modifying the modulepath
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-   If you have set a module path explicitly in your jobscript, you can remove it from your jobscript or change it to the module path for Rocky 8.
+As discussed in the previous section, the recommended approach to set your
```
moravveji, line 214: Can you please rephrase this sentence, or even better, split it up into two sentences?

source/leuven/genius_2_rocky.rst
```diff
@@ -1,121 +1,267 @@
...
-   $ module load cluster/genius/batch
+.. _impact_on_user_installed_software:
 
-will set the module path for the modules that are applicable for the Genius ``batch`` partition::
+Impact on user-installed software
+---------------------------------
+If you have installed a software package yourself in your own account, and you
+did this on a Genius CentOS 7 node, it must be recompiled on Genius on a node
+with the new OS. This can be done on one of the available test nodes. Please
+request access to the lpt2_rocky8_pilot group.
```
moravveji, line 251: to the ``lpt2_rocky8_pilot `` group (?)

source/leuven/genius_2_rocky.rst
```diff
@@ -1,121 +1,267 @@
...
 
-   $ module load cluster/wice/batch
+Conda environments
+----------------------------
+The Conda environment you installed might need reinstallations. If you already
+have a Conda environment that works on wICE, it also should work on Genius
+after the migration. If you only have a Conda environment working on Genius,
+it is best to create a new Conda installation. In this case, it is recommended
```
moravveji, line 258: An additional remark can also be given regarding `pip` installs. Something along this line:

"If you used ``pip`` to install a software inside your conda environment and ``pip`` has compiled the package from source, you certainly need to recreate your conda environment."

source/leuven/genius_2_rocky.rst
```diff
@@ -1,121 +1,267 @@
...
+to recreate your environment for full compatibility with the new OS. Best practice
+is to choose a new installation folder with explicit mention of the new OS, e.g.::
 
-will set the the module path of the wICE ``batch`` partition. When you do this on the login node you can examine wich modules are available with the regular commands, e.g.::
+   ${VSC_DATA}/miniconda3-rocky
```
moravveji, line 262: Given that Rocky9 is already out and we are sticking to Rocky8, why not adding the OS version to the path, too? It is just an extra character, but can accommodate possible OS upgrades in the future, too.

source/leuven/genius_2_rocky.rst
```diff
@@ -1,121 +1,267 @@
...
 
-   $ module avail
-   $ module spider cp2k
+In order to install miniconda in a new directory you can ::
 
+   bash Miniconda3-latest-Linux-x86_64.sh -b -p $VSC_DATA/miniconda3-rocky
```
moravveji, line 266: the same here with `rocky` to `rocky8`.

moravveji (approved):
Looks good to me; only few minor textual suggestions are given.

moravveji:
Thanks @stevenvdb for your major revision/upgrade to the page.
I am a bit worried about the overlap with the https://docs.vscentrum.be/software/software_stack.html. But I do not know how other sites manage their software stack? Are all sites using cluster modules?

In the longer term, the content of this page needs to find its way to the software stack pages, and we would better have a timeline for it.

Besides all that, this PR is good to go.

source/leuven/genius_2_rocky.rst
jooghe: That would be indeed a  more futureproof choice

source/leuven/genius_2_rocky.rst
jooghe: As discussed in the previous section, the recommended approach to set your $MODULEPATH environment variable is by using the cluster module. This will make modules from the correct software stack are available,.

jooghe (approved):
Looks good te me too. Just added a few minor answers on Ehsans comments. It is indeed the intention of the CUE to re-write the whoe cluster module part into the general documentation, with then only marking what is different at the different VSC sites.

Issue #215, opened 2021-07-16
TurboVNC


gjbex:
At KU Leuven, TurboVNC is no longer used.  I've removed all references to KU Leuven from pages on which it was mentioned.

I note that UAntwerp has its own page on its visualization nodes.

Do we still need to keep the following page?
./access/turbovnc_start_guide.rst

smoors:
at VUB there is no TurboVNC, only TigerVNC, and we have our own page for it.
so for us it does not matter.

sbecuwe:
If it's no longer used / available at KU Leuven, I suggest to remove it. Don't know whether the slides at the bottom of the page should be kept somewhere? 
Once the page is removed, I'll remove the last sentence from the first paragraph on the UAntwerp page.
We'll need to update ./access/linux_client.rst, ./access/windows_client.rst, ./access/macos_client.rst and ./access/multiplatform_client_tools.rst. The latter page is a bit tricky, since we'll have no longer a "general" TurboVNC page... We could refer to the VNC sections on the linux, windows and macos pages instead?

jooghe:
As this is no longer valid I've tried cleaning up a bit  in /hpcleuven/VscDocumentation/tree/remove_TurboVNC

PR #359, opened 2023-10-06
removing obsolete references to turboVNC at KU Leuven

jooghe:
to solve #215 I removed our obsolete TurboVNC information.


lexming (approved):
I agree to remove this, now we have better solutions with NX and OoD.

backelj (approved):
lgtm

PR #361, opened 2023-10-12
Remove note about Rocky8 pilot group

stevenvdb:
There was still a reference to the `lpt2_rocky8_pilot group`, which is no longer relevant. This PR removes the reference.


moravveji (approved):
LGTM

mselwa (approved):

PR #372, opened 2023-11-29
dont ssh into compute nodes

smoors:
this section is rewritten to use srun instead of ssh to login to compute nodes for 2 reasons:
- using srun keeps the user within the resource constraints of the job, which avoids interference with other jobs, and is more secure, because the user cannot see other jobs (if this is properly setup).
- in Hydra, we've completely disabled ssh access to the compute nodes, because of the security implications.


smoors:
@backelj i've moved all `salloc` sections into the advanced_jobs page (mostly move 'as is').
these sections could use an update, as it is now tailored to antwerp, but i suggest doing that in another PR.

can you have another look?

source/jobs/job_advanced.rst
```diff
@@ -168,3 +168,115 @@ on that node group. On Vaughan the output is rather boring as all nodes are iden
...
+3. We can now execute X11 commands, launch graphical applications, or anything
+   else that we want to do and is supported on a compute node.
+
+4. The first ``exit`` command leaves the compute node and returns to the login
+   shell, **but still within the salloc command**.
+
+5. Hence we need a second ``exit`` command to leave the shell created by
+   ``salloc`` and free the resources for other users.
+
+**So do not forget that you need to exit two shells to free the resources!**
```
lexming, line 282:
```suggestion
.. warning::
   Do not forget that you need to exit two shells to free the allocated resources!
```

lexming (changes requested):

source/jobs/job_types.rst
```diff
@@ -337,121 +334,5 @@ just a single task. To add support for X11, use the ``--x11`` option before ``--
...
-4. The first ``exit`` command leaves the compute node and returns to the login shell, **but still within the
-   salloc command**.
-
-5. Hence we need a second ``exit`` command to leave the shell created by ``salloc`` and free the resources for
-   other users.
-
-**So do not forget that you need to exit two shells to free the resources!**
-
-
+would allocate 64 (virtual) cores, and the second line starts a simple X11 program, ``xclock``,
```
lexming, line 337:
```suggestion
would allocate 64 cores, and the second line starts a simple X11 program, ``xclock``,
```

PR #380, opened 2024-01-15
Update mfa_login.rst

moravveji:
Users of OnDemand do not require SSH key to login. This needed to be made clear in the docs.

### DIFF
--- a/source/access/mfa_login.rst
+++ b/source/access/mfa_login.rst
@@ -4,14 +4,10 @@ Multi Factor Authentication (MFA)
 =================================
 
 |KUL| Multi Factor Authentication (MFA) is an augmented level of security which, as
-the name suggests, requires multiple steps when authenticating. The following
-two factors are necessary to connect to the KU Leuven clusters:
+the name suggests, requires multiple steps when authenticating.
+To login to the KU Leuven clusters, you only need to possess an active account
+with a VSC-associated university or institution.
 
-- A valid private key (except when connecting via :ref:`Open OnDemand portal<ood_t2_leuven>`)
-- Access to a VSC-associated university/institution account
-
-This approach ensures that if your private key is compromised, the person who
-has unauthorized access to it will still not be able to abuse your VSC account.
 This document provides two methods to connect to the KU Leuven clusters with
 MFA. The first method, *without* an agent, is easier to get started with, but requires
 some repetitive steps with each connection. The second method, *with* an agent,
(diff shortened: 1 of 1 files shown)


w-lampaert (approved):

source/access/mfa_login.rst
```diff
@@ -7,7 +7,7 @@ Multi Factor Authentication (MFA)
 the name suggests, requires multiple steps when authenticating. The following
 two factors are necessary to connect to the KU Leuven clusters:
 
-- A valid private key
+- A valid private key (except when connecting via :ref:`Open OnDemand portal<ood_t2_leuven>`)
 - Access to a VSC-associated university/institution account
```
jooghe, line 11: - Access to a VSC-associated university/institution account, with MFA authentication

jooghe (commented):
Just to make sure that the users sees that MFA is also enforced when they are not using a private key.

source/access/mfa_login.rst
stevenvdb: I even think nobody needs a private key anymore, the MFA is entirely within the institution authentication.

moravveji: The only scenario (that I know) where users require keys are NX/FileZilla/WinSCP users, which makes having ssh keys optional for new users.

Is this PR good to go? Or you still propose a change?

jooghe: I think we do indeed need to list the cases when you still need a key. And I think it's best to make the distinction between the KU Leuven Tier-2 and the other VSC infrastructure.
Login in to Hortense with a terminal, you will need a key. However if you use the web interface I'm not sure if you need the key.

moravveji: > > The only scenario (that I know) where users require keys are NX/FileZilla/WinSCP users, which makes having ssh keys optional for new users.
> > Is this PR good to go? Or you still propose a change?
> 
> Can you try to log in using for instance NX, without a key? For me that seems to work, I get a certificate without ever needing my key and that certificate suffices to connect with NX. This makes me think that the private SSH key is never used when connecting. Unless I am wrong about that (and maybe it is good to check with Peter), I would propose to remove the line `A valid private key (except...)` completely.

I can always login to NX with my certificate. but, I think with e.g. FileZilla, you'd always need a key. If we ditch FileZilla/WinSCP in favor of globus, then of course, we do not require keys at all.

lexming:
Due to the changes in https://github.com/hpcleuven/VscDocumentation/pull/452, this file has moved to `source/accounts/mfa_login.rst`.

PR #382, opened 2024-01-23
Retire environment.yaml and other Python env refinements

MaximeVdB:
Currently there are two different files that list the Python packages needed for local rendering:
- `requirements.txt`: used in the ReadTheDocs build and mentioned in the README for Python venvs
- `environment.yaml`: mentioned in the README for Conda environments

`requirements.txt` has been receiving various updates, while `environment.yaml` hasn't and so it got quite outdated. Rather than trying to maintain both, I propose to just stick to `requirements.txt`. 

I then also took the opportunity to restructure the `Prerequisites` subsection a bit and to say something about which Python version one should preferably (but not necessarily) use.


lexming (approved):
LGTM

gjbex (approved):

PR #383, opened 2024-01-24
Update KU Leuven network drive GUI instructions

MaximeVdB:
The migration to Rocky8 also involved an update of Thunar, with changes to its user interface.

(There's also another file manager (Files/Nautilus) where the procedure is quite similar. But to keep it simple this just focuses on the default file manager.)


mselwa (approved):
Works ok here. Thank you.

avapirev (approved):

PR #392, opened 2024-02-16
Add wICE extension hardware details

MaximeVdB:
This just adds the hardware details for the wICE extension, in a separate subsection.

(Once it gets opened for general usage we can put everything in the same list and update the source/leuven/tier2_hardware/wice_hardware/wice.png figure.)

### DIFF
--- a/source/leuven/tier2_hardware/wice_hardware.rst
+++ b/source/leuven/tier2_hardware/wice_hardware.rst
@@ -13,30 +13,30 @@ Hardware details
 - 172 thin nodes 
    
   - 2 Intel Xeon Platinum 8360Y CPUs\@2.4 GHz (Ice lake), 36 cores each
-  - 256 GB RAM 
+  - 256 GiB RAM
   - 960 GB SSD local disk
   - partitions ``batch/batch_long``, :ref:`submit options <submit to wice compute node>`
 
 - 5 big memory nodes
 
   - 2 Intel Xeon Platinum 8360Y CPUs\@2.4 GHz (Ice lake), 36 cores each
-  - 2048 GB RAM
+  - 2048 GiB RAM
   - 960 GB SSD local disk
   - partition ``bigmem``, :ref:`submit options <submit to wice big memory node>`
 
 - 4 GPU nodes, 16 GPU devices
 
   - 2 Intel Xeon Platinum 8360Y CPUs\@2.4 GHz (Ice lake), 36 cores each
-  - 512 GB RAM
-  - 4 NVIDIA A100 SXM4, 80 GB GDDR, connected with NVLink
+  - 512 GiB RAM
+  - 4 NVIDIA A100 SXM4, 80 GiB GDDR, connected with NVLink
   - 960 GB SSD local disk
   - partition ``gpu``, :ref:`submit options <submit to wice GPU node>`
 
 - 5 interactive nodes
 
   - 2 Intel Xeon Gold 8358 CPUs\@2.6 GHz (Ice lake), 32 cores each
-  - 512 GB RAM
-  - 1 NVIDIA A100, 80 GB GDDR 
+  - 512 GiB RAM
+  - 1 NVIDIA A100, 80 GiB GDDR
   - 960 GB SSD local disk
   - partition ``interactive``, :ref:`submit options <submit to wice interactive node>`
 
@@ -56,7 +56,7 @@ which will be made accessible during the spring of 2024:
   - 2 Intel Xeon Platinum 8468 CPUs (Sapphire Rapids),
     48 cores each :raw-html:`<br />`
     The base and max CPU frequencies are 2.1 GHz and 3.8 GHz, respectively.
-  - 256 GB RAM
+  - 256 GiB RAM
   - 960 GB SSD local disk
   - partitions ``batch_sapphirerapids/batch_sapphirerapids_long``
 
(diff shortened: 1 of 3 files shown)


source/leuven/tier2_hardware/wice_hardware.rst
```diff
@@ -44,3 +44,41 @@ The nodes are connected using an Infiniband HDR-100 network, the islands are ind
...
+  - 256 GB RAM
+  - 960 GB SSD local disk
+  - partitions ``batch_sapphirerapids/batch_sapphirerapids_long``
+
+- 4 GPU nodes, 16 GPU devices
+
+  - 2 AMD EPYC 9334 CPUs (Genoa),
+    32 cores each :raw-html:`<br />`
+    The base and max CPU frequencies are 2.7 GHz and 3.9 GHz, respectively.
+  - 770 GB RAM
```
stevenvdb, line 68: Slightly pedantic, but should this not be 768 GiB (=24 * 32 GiB)? Not entirely sure about the GiB vs GB.

source/leuven/tier2_hardware/wice_hardware.rst
```diff
@@ -44,3 +44,41 @@ The nodes are connected using an Infiniband HDR-100 network, the islands are ind
...
+We are currently installing and testing additional hardware for wICE,
+which will be made accessible during the spring of 2024:
+
+- 68 thin nodes
+
+  - 2 Intel Xeon Platinum 8468 CPUs (Sapphire Rapids),
+    48 cores each :raw-html:`<br />`
+    The base and max CPU frequencies are 2.1 GHz and 3.8 GHz, respectively.
+  - 256 GB RAM
+  - 960 GB SSD local disk
```
stevenvdb, line 60: I think we expose only a portion of this to the user, but of course you can argue that these are the correct hardware specs and a limit for users belongs rather in the quick start guide (also see our internal note https://scm.icts.kuleuven.be/issues/83697#note-8).

source/leuven/tier2_hardware/wice_hardware.rst
```diff
@@ -44,3 +44,41 @@ The nodes are connected using an Infiniband HDR-100 network, the islands are ind
 
 .. figure:: wice_hardware/wice.png
```
stevenvdb, line 45: Perhaps the diagram should be extended as well, maybe there is already an Excel showing the physical location of the new nodes.

source/leuven/tier2_hardware/wice_hardware.rst
MaximeVdB: I was basing it on `lsmem` output, but it indeed looks like that cannot always be thrusted (conflicting with what you'd get by counting the DIMMs and with other commands like `lshw`). And all these numbers should be GiB, not GB. I fixed this one and also the entries in our other hardware detail pages (41cf442b6f1d814bd96c65b769dc911d174f1baa).

(Side note: `nvidia-smi` actually reports 79.65 GiB and not 80 GiB for our H100 nodes, but so I've rounded this up to 80 GiB)

source/leuven/tier2_hardware/wice_hardware.rst
MaximeVdB: We indeed consider the hardware detail pages to just be the 'raw' specs.

source/leuven/tier2_hardware/wice_hardware.rst
MaximeVdB: Yes we should do that eventually, but I propose to do that when going into production (as mentioned at the top)

stevenvdb (approved):
Thanks, looks ok to me.

source/leuven/tier2_hardware/wice_hardware.rst
```diff
@@ -13,34 +13,72 @@ Hardware details
...
 The nodes are connected using an Infiniband HDR-100 network, the islands are indicated on the diagram below.
 
 .. figure:: wice_hardware/wice.png
    :alt: wICE hardware diagram
+
+
+Hardware details (extension)
+----------------------------
+We are currently installing and testing additional hardware for wICE,
+which will be made accessible during the spring of 2024:
```
jooghe, line 52: This will be our first Direct Liquid Cooled system.

jooghe: Forget it. It is mentioned below.

source/leuven/tier2_hardware/wice_hardware.rst
```diff
@@ -13,34 +13,72 @@ Hardware details
...
+
+  - 2 Intel Xeon Platinum 8360Y CPUs (Ice lake),
+    36 cores each :raw-html:`<br />`
+    The base and max CPU frequencies are 2.4 GHz and 3.5 GHz, respectively.
+  - 8 TiB RAM
+  - 960 GB SSD local disk
+  - partition ``hugemem``
+
+The thin nodes and GPU nodes are all in the same Infiniband HDR-100 network
+island. These nodes are furthermore the first ones in the data center
```
jooghe, line 83: The GPU nodes are not in the infiniband HDR compute network. They only have the storage IB interconnect.

jooghe (approved):
Only changes my remark concerning the GPU Infiniband.

source/leuven/tier2_hardware/wice_hardware.rst
MaximeVdB: Ah yes; I now fixed this in [1d518ad](https://github.com/hpcleuven/VscDocumentation/pull/392/commits/1d518adc631b6270274b230fc580adfb714b2580)

PR #394, opened 2024-03-01
wICE quickstart & software installation update (wICE extension)

MaximeVdB:
This adds the new partitions to the quickstart (with some additional refinements) and expands the `Compiling software` part of the wICE advanced guide (i.e. what we recommend to users for these new partitions).

(There are of course a few other improvements one can think of. We might need to revise the `Conda on wICE` part, for example. But if you don't mind I would suggest to keep such things for a future PR.)

### DIFF
--- a/source/leuven/wice_advanced_guide.rst
+++ b/source/leuven/wice_advanced_guide.rst
@@ -95,6 +95,30 @@ available. For more general information on software development on the VSC,
 have a look at this :ref:`overview <software_development>`.
 
 
+.. _wice_memory_hierarchy:
+
+Memory hierarchy
+----------------
+
+When running applications in parallel it is often a good idea to take the
+memory hierarchy into account (for example when pinning MPI processes
+in :ref:`hybrid MPI/OpenMP calculations <hybrid_mpi_openmp_programs>`).
+The nodes in the ``batch`` partition on Genius and wICE are the simpler ones
+with a single NUMA domain and L3 cache per CPU, with the usual core-private
+L1 and L2 caches. Other node types may feature more than one NUMA domain per
+CPU and (in the case of AMD CPUs) more than one L3 cache per CPU.
+The 48 cores in a Sapphire Rapids CPU, for example, share a large L3 cache
+but are organized in 4 groups of 12 cores, each group associated with one
+NUMA domain. For a complete overview, please consult the
+:ref:`genius hardware` and :ref:`wice hardware` pages.
+
+.. note::
+
+    You can also retrieve this information using the ``lstopo-no-graphics``
+    command. When on a compute node, keep in mind that the output will only
+    be complete if all available cores have been allocated to your job.
+
+
 .. _wice_worker:
 
 Worker
--- a/source/leuven/wice_advanced_guide.rst
+++ b/source/leuven/wice_advanced_guide.rst
@@ -45,13 +45,15 @@ Conda environments for different CPU types).
 To let jobs use the correct installation at runtime, you can make use of
 predefined environment variables such as ``${VSC_ARCH_LOCAL}`` (and possibly
 ``${VSC_ARCH_SUFFIX}`` and ``${VSC_INSTITUTE_CLUSTER}``) to organize your
-installations.
+installations (see the examples below).
 
 For software using CPUs, the different installations would be:
 
 - one for SkyLake and CascadeLake CPUs
   :raw-html:`<br />`
-  (``${VSC_ARCH_LOCAL}`` = ``skylake`` or ``cascadelake``)
+  (``${VSC_ARCH_LOCAL}`` = ``skylake`` or ``cascadelake``; :raw-html:`<br />`
+  if needed you can control this for your jobs by e.g. adding a
+  ``--constraint=cascadelake`` Slurm option)
 - one for IceLake CPUs
   :raw-html:`<br />`
   (``${VSC_ARCH_LOCAL}`` = ``icelake``)
@@ -74,10 +76,9 @@ For software which also uses GPUs, this would be:
   :raw-html:`<br />`
   (``${VSC_ARCH_LOCAL}`` = ``zen4`` and ``${VSC_ARCH_SUFFIX}`` = ``-h100``)
 
-Compiling software for a particular CPU/GPU model can be done in an
-interactive job (``srun ...``) submitted to a partition containing
-the target device(s) (see e.g. :ref:`genius hardware` and
-:ref:`wice hardware`).
+Unless mentioned otherwise, the ``${VSC_ARCH_SUFFIX}`` corresponds to an
+empty string. You can check which CPU and GPU models are present in which
+partitions on the :ref:`genius hardware` and :ref:`wice hardware` pages.
 
 Many dependencies you might need are centrally installed. The modules
 that are optimized for wICE are available when the appropriate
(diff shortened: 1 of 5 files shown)


source/leuven/wice_advanced_guide.rst
```diff
@@ -9,15 +9,68 @@ wICE advanced guide
...
-on Genius. Although architectural differences are rather small, it is highly
-recommended to compile a version of your code specifically for wICE. A good approach
-to compile on any cluster (such as wICE) is to launch an interactive job (with the
-``srun`` command).
-
-Many dependencies you might need are centrally installed. The modules that are
-optimized for wICE are available when the appropriate
+Compared to the SkyLake and CascadeLake CPUs on Genius, the wICE nodes
+feature more recent CPU models such as Intel IceLake, Intel Sapphire Rapids
+and AMD Genoa. While architectural differences between SkyLake and CascadeLake
```
stevenvdb, line 14: I think this section and the next one can be expanded a bit. Of course it will be a balancing act to provide all useful information without overwhelming the users. Here are two points I would like to add:
* All these CPUs are based on the same instruction set architecture (ISA), but differences arise because different CPUs add additional instructions to the common set. Whether or not a code can benefit from the additional instructions offered by a CPU, is case-specific and cannot be discussed here in general. One very important remark is that the compiler needs to support CPU-specific instructions, which is why choosing a recent compiler on recent hardware is crucial. As an example, the Advanced Matrix Extensions introduced in Sapphire Rapids are only supported by GCC starting from version 11. When using an older compiler (for instance GCC 10.3.0 included in `foss-2021a`), the resulting binary might not be optimized for the Sapphire Rapids architecture.
* Next to differences in the ISA, also the organization of memory access has to be taken into account. The Intel IceLake nodes have 2 NUMA domains, while the Intel Sapphire Rapids and AMD Genoa nodes have 8 NUMA domains. This can be important to consider for instance when pinning MPI processes to cores, but again it is not possible to give general advice because this is case-specific.

These remarks should maybe not be included as-is, but if you can add some of it to this page it would be nice I think.

source/leuven/wice_advanced_guide.rst
```diff
@@ -9,15 +9,68 @@ wICE advanced guide
...
+CPUs can be neglected, the differences with newer CPU models are more
+substantial. When it comes to GPUs there are also significant differences in
+the capabilities of P100, V100, A100 and H100 GPUs.
+
+When locally installing software yourself, we therefore recommend to have
+separate installations for the different types of CPUs and (if applicable)
+GPUs on which you intend to run the software. This rule applies most strongly
+to performance-critical code which is compiled from source (and less strongly
+to light-weight tools based on precompiled binaries).
```
stevenvdb, line 23: precompiled binaries or interpreted code such as pure Python scripts).

source/leuven/wice_advanced_guide.rst
```diff
@@ -9,15 +9,68 @@ wICE advanced guide
...
+    optimized binaries as provided by the centrally installed modules and/or
+    by local installations from source. See also :ref:`conda for Python` and
+    :ref:`wice_conda`.
+
+To let jobs use the correct installation at runtime, you can make use of
+predefined environment variables such as ``${VSC_ARCH_LOCAL}`` (and possibly
+``${VSC_ARCH_SUFFIX}`` and ``${VSC_INSTITUTE_CLUSTER}``) to organize your
+installations.
+
+For software using CPUs, the different installations would be:
```
stevenvdb, line 40: I am not sure that the intention here will be clear. Maybe it helps to include a practical example, for instance a job script that shows how to use this?
Maybe it is worth mentioning that on most nodes, `${VSC_ARCH_SUFFIX}` is set to an empty string on most of our nodes?

stevenvdb (commented):
I have added some suggestions.

> We might need to revise the Conda on wICE part, for example. But if you don't mind I would suggest to keep such things for a future PR.

I am not sure; after this PR, there would be a recommendation to have architecture-specific installations, but at the bottom of the page an example of a cluster-specific installation for conda. That might be confusing.
On the other hand, do we really want users duplicating their conda environment several times? Maybe we really need to add a warning to that section that users only need to duplicate environments if:
* the code from the environment requires a lot of total compute time
* that compute time is spent in pre-compiled libraries for which architecture-specific releases are available
* they plan to run jobs on multiple architectures

source/leuven/wice_advanced_guide.rst
```diff
@@ -9,15 +9,68 @@ wICE advanced guide
...
+  :raw-html:`<br />`
+  (``${VSC_ARCH_LOCAL}`` = ``icelake``)
+- one for AMD Genoa CPUs with H100 GPUs
+  :raw-html:`<br />`
+  (``${VSC_ARCH_LOCAL}`` = ``zen4`` and ``${VSC_ARCH_SUFFIX}`` = ``-h100``)
+
+Compiling software for a particular CPU/GPU model can be done in an
+interactive job (``srun ...``) submitted to a partition containing
+the target device(s) (see e.g. :ref:`genius hardware` and
+:ref:`wice hardware`).
```
jooghe, line 70: target CPU's/GPU's instead of devices?

source/leuven/wice_quick_start.rst
```diff
@@ -50,109 +50,108 @@ more information on the following pages:
...
 
 Submit to a GPU node
 ~~~~~~~~~~~~~~~~~~~~
 
-The GPU nodes are located in the ``gpu`` partition, so you will need to explicitly specify 
-it when submitting your job. 
-Similar to the other nodes, the GPU nodes can be shared by different jobs from different users.
-However every user will have exclusive access to the number of GPUs requested. 
-If you want to use only 1 GPU of type A100 you can submit for example like this::
+The nodes with A100 GPUs are located in the ``gpu`` partition. As for the other
```
jooghe, line 138: will we here already mention the gpu_a100 partition?

source/leuven/wice_quick_start.rst
```diff
@@ -50,109 +50,108 @@ more information on the following pages:
...
-  
-Note that in case of 1 GPU you have to request 18 cores. 
-In case you need more GPUs you have to multiply the 18 cores with the number of GPUs 
-requested, so in case of for example 3 GPUs you will have to specify this::
+   $ sbatch --account=lp_myproject --clusters=wice --partition=gpu \
+            --nodes=1 --ntasks=54 --gpus-per-node=3 myjobscript.slurm
 
-   $ sbatch --account=lp_myproject --clusters=wice --nodes=1 --ntasks=54 \
-            --gpus-per-node=3 --partition=gpu myjobscript.slurm
+There are also nodes with 4 H100 GPUs (and 64 AMD Genoa CPUs) which you
```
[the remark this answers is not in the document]
MaximeVdB: (addressed in 3ffd8479b6e9e5046e1c5d7dc31884c6e666e56e)

source/leuven/wice_advanced_guide.rst
MaximeVdB: thanks, I've given this a try in d454b75660fc68a5e9f1c36d9c0f17957fc4237d and be5e608b63f43a8a22c3cba6c488ca3cf8c24673

source/leuven/wice_advanced_guide.rst
MaximeVdB: (changed as part of 5bfa41d2c2a706327ab763c179759536d3c6d765)

source/leuven/wice_advanced_guide.rst
MaximeVdB: There was indeed still room for improvement there (-> 5bfa41d2c2a706327ab763c179759536d3c6d765)

source/leuven/wice_advanced_guide.rst
MaximeVdB: (changed as part of 5bfa41d2c2a706327ab763c179759536d3c6d765)

MaximeVdB:
> I have added some suggestions.
> 
> > We might need to revise the Conda on wICE part, for example. But if you don't mind I would suggest to keep such things for a future PR.
> 
> I am not sure; after this PR, there would be a recommendation to have architecture-specific installations, but at the bottom of the page an example of a cluster-specific installation for conda. That might be confusing. On the other hand, do we really want users duplicating their conda environment several times? Maybe we really need to add a warning to that section that users only need to duplicate environments if:
> 
>     * the code from the environment requires a lot of total compute time
> 
>     * that compute time is spent in pre-compiled libraries for which architecture-specific releases are available
> 
>     * they plan to run jobs on multiple architectures

I've now also given an update to the Conda-related parts (-> 32cfe7cb6bd245e4700980e7aa9f2e2bd7fb3064)

source/leuven/wice_quick_start.rst
MaximeVdB: Those ones weren't there yet when I opened the PR, but in the meantime this has been deployed. So I've also added those partition aliases now in b13491a50fdf23618ac71ba8e7ce886d4bd45df5.

stevenvdb (approved):
Thanks for the changes, for me this is good to go.

Only one side note; does anybody know why the bigmem nodes have a different memory configuration compared to the hugemem node?

mselwa (approved):

PR #395, opened 2024-03-08
Feature/antwerp updates march 2024

RVerschoren:
This is an update to reflect the changes made in the summer of 2023.

This mostly touches only UAntwerp documentation, removing old UAntwerp information and updating the hardware pages to reflect the current state.

W.r.t. shared documentation: it replaces the Hopper by our Breniac cluster as Slurm cluster, adds Vaughan to the list of clusters providing GPU access, and updates an old link to the TurboVNC download.


sbecuwe:
- Problem with numbering on hardware-tier2.html? (Correct numbering in navigation pane, but not in main window)
```
2. Tier-2 Infrastructure
2.1. UAntwerp Tier-2 Infrastructure
2.1.7. Login infrastructure
```
- Shouldn't we add Hopper to hardware-archive.html?
- Suggest to run a spell check on the _complete_ website, e.g. visuali[zs]ation

RVerschoren:
> * Problem with numbering on hardware-tier2.html? (Correct numbering in navigation pane, but not in main window)
> 
> 
> ```
> 2. Tier-2 Infrastructure
> 2.1. UAntwerp Tier-2 Infrastructure
> 2.1.7. Login infrastructure
> ```

This was due to the hidden toctree at the top of the file. Placing it at the bottom resolved the numbering.

>     * Shouldn't we add Hopper to hardware-archive.html?

Done

>     * Suggest to run a spell check on the _complete_ website, e.g. visuali[zs]ation

The occurrences now all say "visualization" on our pages, as that is the most prominent spelling across the site. Three source files still use "visualisation" (`access/nx_start_guide.rst`,`software/postprocessing_tools.rst` and `leuven/services/openondemand.rst`), but I suspect it's better to open a separate PR for this.

source/antwerp/tier2_hardware.rst
```diff
@@ -1,12 +1,270 @@
...
+
+      pip install --prefix=${VSC_DATA}/python_lib sphinx
+* |VUB| |UG| |KUL| **Users with a different home institution**
+        
+   VSC_DATA is a file system at your local
+   institution. Due to the distance between our clusters and your home institution, file access to 
+   VSC_DATA will have a high latency. 
+   
+   Installing the packages that you need locally on **VSC_SCRATCH**
+   will give much better performance.
```
lexming, line 223: There is a formatting issue here with the indentation:
```suggestion
* |UA| **Users with a vsc2xxxx account**
   
  When working on the UAntwerp clusters, VSC_DATA is a local file
  system for you. We expect that for :ref:`Python <Python packages>` packages 
  (and similarly for :ref:`R <R_package_management>` and :ref:`Perl <Perl packages>`), 
  **VSC_DATA** will give better performance than VSC_SCRATCH due to the amount of file 
  metadata accesses that occur when running Python.

  |Example| To install the ``sphinx`` Python package, you can use the following command
   
  .. code-block:: bash

     pip install --prefix=${VSC_DATA}/python_lib sphinx

* |VUB| |UG| |KUL| **Users with a different home institution**
        
  VSC_DATA is a file system at your local
  institution. Due to the distance between our clusters and your home institution, file access to 
  VSC_DATA will have a high latency. 
   
  Installing the packages that you need locally on **VSC_SCRATCH**
  will give much better performance.
```

source/antwerp/remote_visualization_uantwerp.rst
```diff
@@ -142,243 +26,240 @@ section for information about ssh clients for various client operating systems.
...
 
-#. When you've finished, don't forget to log out in the Xfce desktop 
+   |Example| Start MATLAB with hardware-accelerated OpenGL with ``-nosoftwareopengl`` as follows:
+
+    .. code:: bash
+ 
+       vglrun matlab -nosoftwareopengl
+    
+    The MATLAB command ``opengl info`` will then show that you are
+    indeed using the GPU.
```
lexming, line 260: There is a formatting issue here with the indentation:
```suggestion
   |Example| Start MATLAB with hardware-accelerated OpenGL with ``-nosoftwareopengl`` as follows:

   .. code:: bash
 
      vglrun matlab -nosoftwareopengl
    
   The MATLAB command ``opengl info`` will then show that you are
   indeed using the GPU.
```

lexming (changes requested):
This update is very nice. I just spotted a few formatting issues. See below

source/antwerp/tier2_hardware.rst
RVerschoren: This extra space in the formatting was added deliberately to get the border on the left side of the blocks.
However, if it's more in line with the rest of the documentation, the suggestion is fine for me too.

source/antwerp/remote_visualization_uantwerp.rst
RVerschoren: This was also indented extra for the bar to the left, keeping the example together.

Is there another way to separate this example cleanly from the rest of the instructions, or would you say this is OK (as it is quite a small example indeed)?

lexming (approved):
LGTM

lexming:
@backelj on my side this is ready

backelj:
The pages have UAntwerp in the title, while the tabs (e.g., on VPN connection) have UAntwerpen — can we use the same name everywhere? Not sure which one to prefer though...

backelj (approved):
lgtm — most likely, some pages will need some extra/more changes, but for the time being, this version can be approved so that it's at least online...

PR #401, opened 2024-04-16
update matlab parallel documentation

w-lampaert:


source/software/matlab_parallel_computing.rst
```diff
@@ -7,117 +7,93 @@ General
...
-    $ git clone https://github.com/hpcleuven/matlab-remote.git
-    $ cd <your_preferred_location>/matlab-remote
 
-Within the folder, you will see a ``mdcs.rc`` file. This is a configuration file that allows you to adapt the cluster profile to your needs. There is only one setting 
-that needs to be changed here: ``LocalJobStorageLocation``. This will define where your job output will be directed to. This script will not create the folder for you, 
-so be sure to create it yourself if it does not exist yet. There is no need for quotation marks when defining this path. 
+    $ module load <matlab_module>
 
-> **_NOTE:_**  Be careful: when leaving this field blank, this location will default to your ``$VSC_HOME``, which could result in quickly filling up your home directory.
+We have Matlab modules available for wICE (in the 2021a toolchain) and for Genius (in the 2021a and 2018a toolchain).
```
moravveji, line 37: ... and 2018a toolchains).

source/software/matlab_parallel_computing.rst
```diff
@@ -7,117 +7,93 @@ General
...
+    $ c.AdditionalProperties
+    $
+    $ #edit
+    $ c.AdditionalProperties.<property_name> = '<property_value>'
+
+One of these properties, you will always have to set, namely `AccountName`. Set this to the name of your credit account.
 
-If you would prefer to use the local resources (i.e. the resources of the node you are currently on), use:
+Next to this, it is also possible to change the more standard parameters of the cluster profile. One of these is the `JobStorageLocation` parameter. This defaults
+to a location in your `$VSC_HOME`. If you are producing a lot of output, it might be good to change this location to e.g. `$VSC_DATA`. Changing this, or any of
```
moravveji, line 68: normally users shall avoid using their home folder in all cases. So, please reformulate like:
"It is strongly recommended to change the default storage location to ..."

moravveji (changes requested):
Because I am unfamiliar with Matlab, I am only giving textual suggestions here.

moravveji (approved):
lgtm

source/software/matlab_parallel_computing.rst
```diff
@@ -129,43 +105,15 @@ You can start an interactive job using the ``parpool`` function:
 ::
 
     >> c = parcluster;
-    >> p = parpool(72); % requesting 72 cores
```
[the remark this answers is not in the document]
w-lampaert: That is somewhat more complicated as it is something that Matlab determines on itself. Normally you would think that it just request 1 node and up to 72 cores with parpool(<=72) , but I have seen other scenarios where there are multiple nodes even when there are less than 72 cores. I suspect it just does a more general request, and receives whatever is available.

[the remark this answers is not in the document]
jan-de-laet: But 'whatever is available' is determined by the srun/salloc command that is used, no? Assume for the sake of the argument that a user wants to use a parpool of 2. Here are two srun commands that could be used (executed on the sandbox):

$ srun --nodes=2   -A lpt2_sysadmin --pty bash -l
srun: job 60004811 queued and waiting for resources
srun: job 60004811 has been allocated resources
[...]
SLURM_NNODES: 2
SLURM_NODELIST: e25c03n[1-2]
SLURM_JOB_CPUS_PER_NODE: 1(x2)

versus

$ srun --nodes=1 --ntasks=2   -A lpt2_sysadmin --pty bash -l
[]
SLURM_NNODES: 1
SLURM_NODELIST: e25c03n1
SLURM_JOB_CPUS_PER_NODE: 2
Date: Tue Apr 30 08:30:43 CEST 2024
Walltime: 00-01:00:00

In both cases Matlab has two cpus at its disposal (I would think), but I would assume that having two cores on one node is better than having one core on each of two nodes. 

In a way my question is more general than parallel matlab, and this indeed is not the place to educate the users on that. But perhaps it may be useful to have a small comment that the matlab parpool uses the resources available, and that these available resources are determined by the options of srun/salloc? At least if that is what happens :-).

jan-de-laet (commented):
See comment at line 132

w-lampaert:
I added a note to explain this

w-lampaert:
I did some extra updates after revisiting the installation protocol as well.

source/software/matlab_parallel_computing.rst
```diff
@@ -7,119 +7,99 @@ General
...
 
 ::
 
-   $ # Genius
-   $ module use /apps/leuven/skylake/2021a/modules/all
-   $ module load MATLAB/R2022b #MATLAB/R2022a if you would like to load this version
-    
-Once the module is loaded, start Matlab. You can start it in a terminal without the GUI by using the ``-nodisplay`` flag. Then, run the ``configCluster`` script.
-This will set up the parallel configuration.  
+    $ matlab -nodisplay          # the '-nodisplay' option avoids launching the GUI, but you leave this out if you prefer (slow)
```
moravveji, line 44: ... but you may leave this out (slow)

source/software/matlab_parallel_computing.rst
```diff
@@ -7,119 +7,99 @@ General
...
+    $ # Within Matlab
+    $ >> configCluster;
 
-::
+The `configCluster` function will request you to specify a cluster (`wice` or `genius`). Please use the one you are on already.
```
moravveji, line 48: Please choose the one ...

source/software/matlab_parallel_computing.rst
```diff
@@ -7,119 +7,99 @@ General
...
+    $ # Access
+    $ c.AdditionalProperties
+    $
+    $ #edit
+    $ c.AdditionalProperties.<property_name> = '<property_value>'
 
-If you would prefer to use the local resources (i.e. the resources of the node you are currently on), use:
+One of these properties you will always have to set, namely `AccountName`. Set this to the name of your credit account.
+
+Next to this, it is also possible to change the more standard parameters of the cluster profile. One of these is the `JobStorageLocation` parameter. This defaults
```
moravveji, line 69: \`JobStorageLocation\` -> \`\`JobStorageLocation\`\`

source/software/matlab_parallel_computing.rst
```diff
@@ -7,119 +7,99 @@ General
...
+to a location in your `$VSC_HOME`. It is however strongly recommended to change this to another location, e.g. `$VSC_DATA`. Changing this, or any of
```
moravveji, line 70: the same as above for VSC_{HOME,DATA}

source/software/matlab_parallel_computing.rst
```diff
@@ -7,119 +7,99 @@ General
...
+Both after changing any of the additional properties or the standard parameters, you will have to save that profile to keep the changes after closing your
+session:
 
 ::
 
-   >> % View the additional properties
-   >> c.AdditionalProperties
-   >> % set the required properties
-   >> c.AdditionalProperties.AccountName = <account-name>;
-   >> c.AdditionalProperties.WallTime = '05:00:00';
```
moravveji, line 102: Specifying `WallTime` is also an interesting argument. Also, if it is possible to specify the number of nodes and cores per node and memory, if the profile allows?

source/software/matlab_parallel_computing.rst
```diff
@@ -7,119 +7,99 @@ General
...
+
 
 Submitting jobs
 +++++++++++++++
 
 Submitting jobs will happen from within Matlab. There are two ways to do this. You can start a Matlab session on a login node, or you can first request an
 interactive job, where you launch your Matlab session. Which one you choose will depend on the workload you plan to have on the node before submitting a job.
-As for any other work on the cluster, you can always do minor work on the login node, but for heavier calculations (before submitting your multinode job), you should
-use an interactive job. 
+As for any other work on the cluster, you can always do minor work on the login node, but for heavier calculations (before submitting your effective job), you should
```
moravveji, line 91: this sentence contradicts the first sentence of the paragraph.
What shall the users eventually do? Can they use the login node for this or not?
If not, the first sentence needs a polish too.

source/software/matlab_parallel_computing.rst
```diff
@@ -7,119 +7,99 @@ General
...
 is best to do this via a NoMachine connection. If you then need to launch the Matlab GUI in an
-interactive job, don't forget to add ``-X`` option to qsub and keep in mind that the GUI will
-necessarily be less responsive.
+interactive job, don't forget to add ``-X`` option to your `srun` or `salloc` command and keep in mind
+that the GUI will necessarily be less responsive.
 
-Now you can start both interactive and independent batch jobs with the previously configured cluster profile. Follow the steps below to submit a job.
+Now you can start both interactive and independent batch jobs with the previously configured cluster profile. In both cases, you request a number of threads, either using the `parpool` or `batch`
+function. Under the hood, Matlab launches a Slurm job to request resources of the system. Different from submitting a Slurm job yourself, there is no direct way to specify the number of nodes and cores.
+Matlab just requests a number of tasks, and the system gives whatever is available, meaning that cores could be spread out over a number of nodes.
```
moravveji, line 102: the good thing is that the cluster configuration file uses `salloc` with the `-N 1` option, right?
If so, then users are bound to as many cores available on a single node, but they are guaranteed to stay within a node!
is that correct?

moravveji (changes requested):

source/software/matlab_parallel_computing.rst
w-lampaert: I think the ``salloc`` statement we found was not related to how Matlab submits the jobs eventually, that was a testing step for the connection I think.

Anyway, it is also possible to request more cores than the max on a node, so I don't think Matlab submits with ``-N 1``

moravveji (approved):

source/software/matlab_parallel_computing.rst
jan-de-laet: "Under the hood, Matlab launches a Slurm job to request resources of the system. Different from submitting a Slurm job yourself, there is no direct way to specify the number of nodes and cores. Matlab just requests a number of tasks, and the system gives whatever is available, meaning that cores could be spread out over a number of nodes."

But if this is done on an interactive node, the "whatever is available" is determined by the nodes and cores specified for that initial interactive slurm job, no? If that is not correct, can you explain why (because I seem to be missing some basic understandig of this then). If it is correct, I still think it would be useful to add something about that (see my example in my previous comment).

source/software/matlab_parallel_computing.rst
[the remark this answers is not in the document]
w-lampaert: It can be both scenarios. I am pretty sure that this is caused by a not so good implementation by Mathworks, and I am not so keen on complicating this matter for our users by trying to explain this with what happens in the background. I completely agree that this is messy, but I think it is better that we take this up with Mathworks and let them revise our parallel server scripts (which are partially custom-made for us by them).

source/software/matlab_parallel_computing.rst
w-lampaert: No, it is not limited by the original job. The logic is like this: 
1. You start an interactive job with whatever amount of resources you want (enough decently start Matlab and then perhaps perform some minor computing, e.g. preparation steps for your Matlab job)
2. You open Matlab
3. You use Matlab to submit a new job to the cluster: You use the parpool or batch function to either send a 'srun/salloc' or a 'sbatch' command to the cluster. Just as in any interactive session, you can submit a job to other nodes from there, the difference here is that you use a Matlab function to assemble your proper slurm command instead of typing it in directly. On top of that, the 'parpool' function will connect to the interactive session and allow you to perform Matlab commands there. The 'batch' function will directly submit certain Matlab commands/scripts to run in the background.

jan-de-laet (approved):

PR #404, opened 2024-04-24
Use default user installation prefix for Jupyter kernels

MaximeVdB:
~~I suppose that `${VSC_HOME}/.local` used to be the place at some point in the past, but now at least it's necessary to use `${VSC_DATA}/.local/`, otherwise the installed kernels won't show up in the web client.~~

With these modifications the kernels will be installed in an appropriate location (searched by the Jupyter server), also when the user e.g. has set the `${XDG_DATA_HOME}` variable.


w-lampaert:
Both commands work as expected for me

WouterVanAssche (approved):

PR #410, opened 2024-05-13
Update superdome_quick_start.rst

moravveji:
112 cores = 8 sockets x 14 cores/socket. I guess by typo, `--ntasks=28` was mentioned.

### DIFF
--- a/source/leuven/superdome_quick_start.rst
+++ b/source/leuven/superdome_quick_start.rst
@@ -37,10 +37,10 @@ job will therefore get two sockets (and so 28 cores in total)::
            --ntasks=2 myscript.slurm
 
 If you want to get multiple Slurm tasks per socket you will need to use the
-``--ntasks-per-socket`` option, for example::
+``--ntasks-per-socket`` option. For example, to exploit all 8 sockets on the node::
 
   $ sbatch --account=lp_my_project --clusters=genius --partition=superdome \
-           --ntasks=28 --ntasks-per-socket=14 myscript.slurm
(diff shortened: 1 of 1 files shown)


avapirev (approved):

MaximeVdB:
> 112 cores = 8 sockets x 14 cores/socket. I guess by typo, `--ntasks=28` was mentioned.

The original text just said `If you want to get multiple Slurm tasks per socket you will need to use the --ntasks-per-socket option, for example: ...`. So the example just illustrated how to get multiple tasks per socket, in this case with an example involving two sockets (not 8), following on the preceding example (with just one task per socket). So as far as I can see the original text is correct (and in fact the suggestion to use `--ntasks=8 --ntasks-per-socket=14` would be weird because `ntasks` is not divisible by `ntasks-per-socket`).

moravveji:
> > 112 cores = 8 sockets x 14 cores/socket. I guess by typo, `--ntasks=28` was mentioned.
> 
> The original text just said `If you want to get multiple Slurm tasks per socket you will need to use the --ntasks-per-socket option, for example: ...`. So the example just illustrated how to get multiple tasks per socket, in this case with an example involving two sockets (not 8), following on the preceding example (with just one task per socket). So as far as I can see the original text is correct (and in fact the suggestion to use `--ntasks=8 --ntasks-per-socket=14` would be weird because `ntasks` is not divisible by `ntasks-per-socket`).

agreed.

moravveji:
closing this

PR #411, opened 2024-05-17
Update singularity.rst

avapirev:
Added an example how to create a Conda environment in a container from YAML file


gjbex (approved):

source/software/singularity.rst
```diff
@@ -173,6 +173,38 @@ An important advantage of definition files is that they can easily
 be shared, and improve reproducibility.
 
 
+Conda environment in a definition file
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+One way to have a Conda environment in a container is to create it from
+an existing environment YAML file. If we have a Conda environment exported
+in a YAML format file called, e.g., ``user_conda_environment.yml``, then
```
moravveji, line 181: in a YAML file format called e.g. ``...``

source/software/singularity.rst
```diff
@@ -173,6 +173,38 @@ An important advantage of definition files is that they can easily
 be shared, and improve reproducibility.
 
 
+Conda environment in a definition file
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
```
moravveji, line 178: Before showing how, I would propose to tell the users why then shall care about this block.
What is the motivation for the users to install conda in a container instead of e.g. in their VSC_DATA?
- One possible scenario is for groups who work collectively, and they use an identical conda environment (built from a reference YAML recipe). So, this approach lowers the total storage lost for the whole group only due to duplicate conda installations.
- Similarly, those users who share a single conda environment on staging are sometimes hit by slow response of conda (due to Lustre problems with too many small files), which might seldom exhibit itself

moravveji (changes requested):

source/software/singularity.rst
avapirev: I think that all of the above suggested edits fall under the already existing subsection
https://docs.vscentrum.be/software/singularity.html#when-should-i-use-containers
A concise explanation about why and where one would like to use a container. That is true especially for the the first bullet point. Regarding the second bullet point, I think that:
1) sharing environment falls under the "when-should-i-use-containers" section. One could propose to expand that one instead rather than adding new information on the why-and-when container use cases.
2) The slow Lustre is our problem and we need to solve that and not advertise container use as a workaround.

moravveji: This is a matter of choice, and I leave it to you to decide.

moravveji:
I have another remark.
There is a full-fledged page about conda here (https://docs.vscentrum.be/software/python_package_management.html#install-python-packages-using-conda) which overlaps in scope with this PR. I also recommend to make a reference in the conda page (e.g. somewhere after "Alternatives to conda" section) pointing at your new section. Then, users know that this is yet another alternative to the proposed solutions, but elaborated elsewhere.

source/software/singularity.rst
avapirev: Added some changes. Requesting re-review

source/software/singularity.rst
```diff
@@ -173,6 +173,42 @@ An important advantage of definition files is that they can easily
 be shared, and improve reproducibility.
 
 
+Conda environment in a definition file
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+`Conda environments
+<https://docs.vscentrum.be/software/python_package_management.html#install-python-packages-using-conda>`_
+are a convinient solution when it comes to handling own Python-dependent
```
moravveji, line 180: ... are convenient solutions when ...

source/software/singularity.rst
```diff
@@ -173,6 +173,42 @@ An important advantage of definition files is that they can easily
...
+software installations. Having a containerized Conda environment is often
```
moravveji, line 181: in the middle of sentences, "Conda" shall be changed to "conda". This is also compatible with the other page about conda, too (see software/python_package_management.rst)

moravveji (changes requested):

source/software/singularity.rst
avapirev: We could be either compatible, or follow the official name of the software which is Conda.

moravveji: I agree with you. In its official documentation, Conda is mention as "Conda". So, the other page shall be tailored later.

moravveji (approved):
lgtm

gjbex (approved):

PR #420, opened 2024-06-19
OOD update

moravveji:
In this PR, the OOD page is updated concerning the following items:
- new Globus button
- updates about JupyterLab and using Conda kernels
- MATLAB app
- ParaView app

Remarks:
- This PR does NOT include any update for RStudio Server; this shall happen once the issues with 'RStudio as a module' are solved.

ToDo List:
- [x] find the oldest version of ipykernel which supports the `--env` option
- [x] can we see/log the kernel usage on JupyterLab?
       answer is: No

### DIFF
--- a/source/leuven/services/openondemand.rst
+++ b/source/leuven/services/openondemand.rst
@@ -21,23 +21,28 @@ Use
 
 The KU Leuven Open OnDemand page provides a range of functions:
 
-- Exploring, adapting and creating files
+- Browsing, creating, transferring, viewing and/or editing files
 - Submitting and monitoring jobs, creating job templates
 - Opening a shell on one of the login nodes
 - Using interactive apps
 
-All of these functionalities can be used by accessing them through the tabs at the top of the page. In the following, we will describe these in
-some more detail.
+All of these functionalities can be used by accessing them through the tabs at the top of the page.
+In the following, we will describe these in some more detail.
 
 Files
 =====
 
-This menu provides a file explorer that allows you to navigate files by clicking. You can access your ``$VSC_HOME`` and ``$VSC_DATA`` folders. Other storage is not
-available here. General file explorer options like moving, deleting, modifying and creating files or directories are available as well. You can also use this interface
-to download and upload files to and from your local machine. Be aware that this is not recommended for very large files.
+This menu provides a file explorer that allows you to navigate through your files and folders.
+You can access your ``$VSC_HOME`` and ``$VSC_DATA`` folders. Other storages are not available here.
+General file explorer options like moving, deleting, modifying and creating files or directories are available as well.
+You can also use this interface to download and upload files to and from your local machine. Be aware that this is not recommended for very large files.
 
 **Good to know:** the standard 'ctrl+s' does not save your edited files on Open OnDemand, but will trigger a save on your local machine. Luckily, there is a 
 save button in the upper left corner on the editor page.
+
+The ``Globus`` button takes you directly to the Globus login page, and upon a successful login to your Globus account
+(using your KU Leuven credentials), you will land at the same sub-directory from which you clicked on the Globus button.
+For more information about Globus, please refer to :ref:`Globus Platform documentation<globus platform>`_.
 
 Jobs
 ====
--- a/source/leuven/services/openondemand.rst
+++ b/source/leuven/services/openondemand.rst
@@ -338,12 +338,13 @@ If you are interested in this approach, follow these steps to create and use you
 - Pick a specific 'Toolchain and Python versions', e.g. '2023a and ``Python/3.11.3-GCCcore-12.3.0``'.
 - Choose a specific architecture, e.g. Sapphire Rapids nodes on wICE
 - Start an :ref:`Interactive Shell<interactive_shell>` on the targeted architecture,
-  and execute the following to create an environment called ``venv_science``:
+  and execute the following to create an environment called ``<kernel_name>``; you may choose a ``<kernel_name>``
+  which represents your project scope:
 
   .. code-block :: bash
 
      TOOLCHAIN='2023a'
-     DIR_VENV=${VSC_DATA}/apps/${VSC_OS_LOCAL}/${VSC_ARCH_LOCAL}${VSC_ARCH_SUFFIX}/${TOOLCHAIN}/venv_science
+     DIR_VENV=${VSC_DATA}/apps/venv/${VSC_OS_LOCAL}/${VSC_ARCH_LOCAL}${VSC_ARCH_SUFFIX}/${TOOLCHAIN}/<kernel_name>
      mkdir -p ${DIR_VENV}
      # the line below is needed if you use 'Interactive Shell' app
      module use /apps/leuven/${VSC_OS_LOCAL}/${VSC_ARCH_LOCAL}${VSC_ARCH_SUFFIX}/${TOOLCHAIN}/modules/all
@@ -352,7 +353,7 @@ If you are interested in this approach, follow these steps to create and use you
      source ${DIR_VENV}/bin/activate
      pip install --prefix=${DIR_VENV} ipykernel <additional packages>
      # note that below, the "--env ..." argument is not needed
-     python -m ipykernel install --user --name venv_science --display-name venv_science
+     python -m ipykernel install --user --name <kernel_name> --display-name <kernel_name>
      # you need this below
      echo ${DIR_VENV}
 
--- a/source/leuven/services/openondemand.rst
+++ b/source/leuven/services/openondemand.rst
@@ -269,12 +269,14 @@ which we explain below.
 Pure module environment
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-In the app resource form, besides the normal choices (:ref:`listed above <interactive-apps>`), you can also choose
-a 'Toolchain version' such as '2023a' from a drop-down menu.
-Based on that choice, the corresponding JupyterLab module will be loaded together with its dependencies (specifically Python).
+In the app resource form, besides the normal choices (:ref:`listed above <interactive-apps>`),
+you can also choose from different 'Toolchain and Python versions' from a drop-down menu.
+An example would be '2023a and ``Python/3.11.3-GCCcore-12.3.0``'.
+Based on that choice, the corresponding JupyterLab module will be loaded together with its
+dependencies (specifically Python).
 
 Furthermore, you may optionally choose to load ``SciPy-bundle`` (for widely-used packages like ``scipy``,
-``numpy``, ``pandas`` and more) and/or ``matplotlib`` in your environment from the same 'Toolchain version'.
+``numpy``, ``pandas`` and more) and/or ``matplotlib`` in your environment from the same toolchain.
 
 Once you launch a JupyterLab session, a default kernel called ``Python 3 (ipykernel)`` is already available in your session.
 This kernel, in addition to the Python standard library, would enable using extra packages from
@@ -283,7 +285,7 @@ This kernel, in addition to the Python standard library, would enable using extr
 .. warning::
 
    If you use JupyterLab as explained above, and you need to reproduce your numerical results 
-   at any time in the future, make sure you stick to choosing the same 'Toolchain version'.
+   at any time in the future, make sure you stick to choosing the same toolchain.
 
 If the module-based environment explained above does not provide all packages that you need, then
 it is recommended to manage your custom-made R or Python environments by creating custom Jupyter kernels.
@@ -296,8 +298,7 @@ If you have not installed Conda in your account yet, please refer to the
 
 To create a custom kernel, first create a  :ref:`Python <create_python_conda_env>` or
 :ref:`R <create_r_conda_env>` Conda environment. The second step consists of effectively
-creating the kernel in your ``$VSC_HOME/.local/share`` folder (which is a defalt value for the
-``$XDG_DATA_HOME`` environment variable), as Jupyter will look for kernels in this location.
+creating the kernel. 
 The following commands should be excecuted from a shell (e.g. using 'Login Server Shell Access'), and only need
 to be done once for the set-up of each new kernel.
 If you already have an existing Python kernel, but your JupyterLab session freezes/craches when choosing your
@@ -325,30 +326,29 @@ virtual environments are architecture specific.
 Hence, on our current Tier-2 machines, this approach needs to be handled with care.
 If you are interested in this approach, follow these steps to create and use your kernel:
 
-- Pick a specific Python from a specific 'Toolchain version', e.g. ``Python/3.11.3-GCCcore-12.3.0``
-  from ``2023a``
+- Pick a specific 'Toolchain and Python versions', e.g. '2023a and ``Python/3.11.3-GCCcore-12.3.0``'.
 - Choose a specific architecture, e.g. Sapphire Rapids nodes on wICE
 - Start an :ref:`Interactive Shell<interactive_shell>` on the targeted architecture,
-  and execute the following:
+  and execute the following to create an environment called ``venv_science``:
 
   .. code-block :: bash
 
      TOOLCHAIN='2023a'
-     DIR_VENV=${VSC_DATA}/${VSC_OS_LOCAL}/${VSC_ARCH_LOCAL}${VSC_ARCH_SUFFIX}/${TOOLCHAIN}
+     DIR_VENV=${VSC_DATA}/apps/${VSC_OS_LOCAL}/${VSC_ARCH_LOCAL}${VSC_ARCH_SUFFIX}/${TOOLCHAIN}/venv_science
      mkdir -p ${DIR_VENV}
+     # the line below is needed if you use 'Interactive Shell' app
      module use /apps/leuven/${VSC_OS_LOCAL}/${VSC_ARCH_LOCAL}${VSC_ARCH_SUFFIX}/${TOOLCHAIN}/modules/all
      module load Python/3.11.3-GCCcore-12.3.0
      python -m venv ${DIR_VENV}
      source ${DIR_VENV}/bin/activate
      pip install --prefix=${DIR_VENV} ipykernel <additional packages>
-     # note that below, --env is not needed
-     python -m ipykernel install --user --name '<env_name>' --display-name '<kernel_name>'
+     # note that below, the "--env ..." argument is not needed
+     python -m ipykernel install --user --name venv_science --display-name venv_science
      # you need this below
      echo ${DIR_VENV}
 
 - On the JupyterLab form, choose a partition that matches your choice of architecture;
-  also, pick the same 'Toolchain version' as above.
-  In the 'Pre-run scriptlet' box, insert ``source <path/pointing/to/DIR_VENV>/bin/activate``.
+  also, pick the same toolchain as above.
 - Once you connect to your session, your new ``<kernel_name>`` is ready to be used.
   To verify your setup, you can execute ``import sys; sys.executable`` in your notebook,
   and the resulting path shall point at ``DIR_VENV`` where you installed your virtual environment.
@@ -363,10 +363,6 @@ With the following command you can create the kernel::
 
 Once the kernel is created, you will see it in the 'Launcher' menu.
 You can now start working in your own customized environment.
-Note that all user kernels are stored by default in ``${VSC_HOME}/.local/share/jupyter/kernels``
-(where ``${VSC_HOME}/.local/share`` is the default value for ``$XDG_DATA_HOME``);
-we strongly advice you to stay away from modifying the contents of this folder, unless you are
-aware of the consequences.
 
 For more general information, please refer to the `official JupyterLab documentation`_.
 
@@ -374,6 +370,21 @@ For more general information, please refer to the `official JupyterLab documenta
 
 - The top-level notebook directory is by default ``$VSC_DATA``.
 - At the moment, we do not support installing extensions in JupyterLab.
+
+.. note::
+
+   User kernels are stored by default in ``${VSC_HOME}/.local/share/jupyter/kernels``.
+   To override this and store your kernel specifications in a non-default location,
+   you may drop the following line in your ``${VSC_HOME}/.bashrc``::
+
+      export XDG_DATA_HOME=${VSC_DATA}/.local/share
+
+   If you have created multiple kernels (for Python and R), you find the corresponding
+   directories under ``${XDG_DATA_HOME}/jupyter/kernels``.
+   To remove a kernel, find and delte the corresponding folder inside the ``kernels``
+   subdirectory.
+   We strongly advice you to stay away from modifying the contents of this folder,
+   unless you are aware of the consequences.
 
 RStudio Server
 --------------
--- a/source/leuven/services/openondemand.rst
+++ b/source/leuven/services/openondemand.rst
@@ -428,8 +428,10 @@ for more detailed information.
 The Tensorboard interactive session requires you to specify a project (or log) directory in 
 your submission options.
 This is a relative directory starting from your ``$VSC_DATA``.
-It is not possible to navigate to the correct folder from within the app.
-
+Beware that you cannot change this directory, once the session is launched.
+If you redirect Tensorboard to a wrong folder (typo in path name or missing log files),
+Tensorboard fails to start, and your session lands on an error page starting with the message:
+'No dashboards are active for the current data set.'.
 
 Code Server
 -----------
--- a/source/leuven/services/openondemand.rst
+++ b/source/leuven/services/openondemand.rst
@@ -195,6 +195,7 @@ Beware that by launching any app you will end up in a regular queue, so requesti
   or defining specific environment variable(s).
   **Beware that this feature is still somewhat experimental, and its functionality also depends on the app you are running
   (mainly RStudio Server has some issues here). If you would like to use this feature, but you run into problems, please contact our helpdesk.**
+
 Once you've selected all your resources, just press 'Launch' and your job will be queued.
 
 .. _choosing_your_resources:
--- a/source/leuven/services/openondemand.rst
+++ b/source/leuven/services/openondemand.rst
@@ -1,8 +1,10 @@
 .. _ood_t2_leuven:
 
-===========================================
 Open OnDemand on the KULeuven Tier2 cluster
 ===========================================
+
+.. sectnum::
+   :depth: 3
 
 About
 =====
--- a/source/leuven/services/openondemand.rst
+++ b/source/leuven/services/openondemand.rst
@@ -327,8 +327,8 @@ Then you create the kernel as follows::
       python -m ipykernel install --user --env PYTHONPATH "" --name '<env_name>' --display-name '<kernel_name>'
 
 
-Custom virtual environments for Python
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Python virtual environments
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 One may create a customized kernel starting from a Python virtual environment, using one of Python modules.
 Such kernels can be created once (e.g. on a login or compute node), and later on be used on any of the
--- a/source/leuven/services/openondemand.rst
+++ b/source/leuven/services/openondemand.rst
@@ -239,47 +239,55 @@ than the shell you get in the "Clusters - Login Server Shell Access" menu, which
 JupyterLab
 -----------
 
-With this app you can use and create Jupyter Notebooks. This can be handy both for R and Python coding.
-In the app resource form, besides the normal choices (:ref:`see above <interactive-apps>`), you can also choose
-a 'Toolchain year' from a drop-down menu, like `2023a`.
+With this app you can create or (re)run Jupyter Notebooks. This can be handy both for R and Python coding.
+In the app resource form, besides the normal choices (:ref:`listed above <interactive-apps>`), you can also choose
+a 'Toolchain year' such as `2023a` from a drop-down menu.
 Based on that choice, a correct JupyterLab module will be loaded together with its dependencies (specifically Python).
 To reproduce your numerical results at any time in the future, make sure you stick to choosing the same 'Toolchain year'.
 
-Furthermore, you may optionally choose to load ``SciPy-bundle`` (for ``scipy``, ``numpy``, ``pandas`` packages and more)
+Furthermore, you may optionally choose to load ``SciPy-bundle`` (for widely-used packages like ``scipy``, ``numpy``, ``pandas`` and more)
 and/or ``matplotlib`` in your environment from the same 'Toolchain year'.
 
-Once you launch a JupyterLab session, you get a default kernel called ``Python 3 (ipykernel)``.
-This kernel, in addition to the Python standard API, would enable using extra packages from from
+Once you launch a JupyterLab session, a default kernel called ``Python 3 (ipykernel)`` is already available in your session.
+This kernel, in addition to the Python standard library, would enable using extra packages from
 ``SciPy-bundle`` and/or ``matplotlib``, if you already selected them.
-However, for using other 
 
-There are two kernels already available, being a Python and a R kernel. The Python and R versions that are used for this, are the versions located in ``/usr/bin``. While you can use these to do some testing, it is not recommended 
-to work with these. It is better to work with conda environments in this case. You should :ref:`install miniconda <install_miniconda_python>` if you have not installed it yet.
-When you do not have any conda environments and their associated kernels, both the Python and R installation will default to the ``~/miniconda3/bin/...`` installation.
+If the "standard" environment explained above does not provide all packages that you need, then
+it is recommended to manage your custom-made R or Python environments using Conda.
+You should :ref:`install miniconda <install_miniconda_python>` if you have not installed it yet.
 
-To create any other kernels, first create a  :ref:`Python <create_python_conda_env>` or 
+To create any other kernel, first create a  :ref:`Python <create_python_conda_env>` or 
 :ref:`R <create_r_conda_env>` conda environment. The second step consists out of effectively
-creating the kernel in your ``$VSC_HOME/.local`` folder, as Jupyter will look for kernels in this location. The following commands should be excecuted from a shell, 
-and only need to be done once for the set-up of each new kernel. This starts with activating your conda
-environment::
+creating the kernel in your ``$VSC_HOME/.local`` folder, as Jupyter will look for kernels in this location.
+The following commands should be excecuted from a shell (e.g. using `Login Server Shell Access'), and only need
+to be done once for the set-up of each new kernel.
+If you already have an existing Python kernel, but your JupyterLab session freeses/craches when choosing your
+old kernel, you also need to repeat the following steps only once.
+This starts with activating your conda environment::
 
       source activate <env_name>
 
-For Python you will need the ``ipykernel`` package installed in your conda environment. Then you create the kernel as follows::
+For Python you will need the ``ipykernel`` package installed in your conda environment::
 
-      python -m ipykernel install --user --name '<env_name>' --display-name '<kernel_name>'
+      conda install ipykernel
+
+Then you create the kernel as follows::
+
+      python -m ipykernel install --user --env PYTHONPATH "" --name '<env_name>' --display-name '<kernel_name>'
 
 For R, you need both the ``jupyter_client`` and the ``irkernel`` package installed. With the following command you can create the kernel::
 
       Rscript -e 'IRkernel::installspec(name="<env_name>", displayname="<kernel_name>")'
 
-Once the kernel is created, you will see it in the 'Launcher' menu. You can now start working in your own customized environment.
+Once the kernel is (re)created, you will see it in the 'Launcher' menu. You can now start working in your own customized environment.
+Note that all user kernels are stored by default in ``${VSC_HOME}/.local/share/jupyter/kernels``, but we strongly advice you
+to stay away from modifying the contents of this folder, unless you are aware of the consequences.
 
 For more general information concerning JupyterLab, go to their `official documentation <https://docs.jupyter.org/en/latest/>`__.
 
 **Remarks:**
 
-- The start location is `$VSC_DATA`
+- The top-level notebook directory is `$VSC_DATA`
 - At the moment, we do not support installing extensions in Jupyter Lab
 
 RStudio Server
--- a/source/leuven/services/openondemand.rst
+++ b/source/leuven/services/openondemand.rst
@@ -259,6 +259,15 @@ from within a Linux terminal.
 This is different than the shell you get in the "Clusters - Login Server Shell Access" menu,
 which directs you towards one of the login nodes.
 
+Currently, the :ref:`cluster modules <cluster_module>` are not automatically loaded when your session starts.
+In order to use modules, one needs to explicitly load the cluster module that adheres to the choice of
+cluster and partition for his or her job.
+For instance, if your job starts on wICE interactive partition, one needs to execute the following command::
+
+    module load cluster/wice/interactive
+
+The same applies for other choices of partitions on Genius or wICE clusters.
+
 JupyterLab
 -----------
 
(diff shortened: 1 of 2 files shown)


moravveji:
@w-lampaert : Before asking other colleagues to review this PR, I'd like you to take a look at changes, specifically the ones concerning JupyterLab, MATLAB and ParaView.

source/leuven/services/openondemand.rst
```diff
@@ -47,7 +52,8 @@ This also means that all your jobs should be submitted as **Slurm** jobs.
 For more detail on how to run jobs on wICE, check out our 
 :ref:`quick start guide<wice_t2_leuven>`.
 
-The jobs tab has two menus, 'Active Jobs' and 'Job Composer':
+The jobs tab has three menus, 'Active Jobs', 'Job Composer' and 'Projects' (which you cannot do much with it
```
w-lampaert, line 55: Perhaps mention that 'projects'  is planned feature for one of the next upstream OOD updates?

source/leuven/services/openondemand.rst
```diff
@@ -141,47 +149,68 @@ the folder that it is associated with.
...
+- Partition: you can choose any of the existing partitions on both clusters.
+  The partition names depend on your choice of cluster.
+  We recommend using the ``interactive`` partition for most interactive work.
 - Numbers of hours: your walltime (min 1h).
 - Number of cores: the amount of cores per node. This defaults to 1.
 - Required memory per core in megabytes. This defaults to 3400 MB.
-- Number of GPUs. Depending on the partition you have requested, you might get a different GPU. The acquired GPU will be the same as the type specified in the partition (e.g. a NVidia V100 for ``gpu_v100``). For wICE, you can also request a GPU from the ``interactive`` partition. One GPU here is a virtual GPU slice of the available A100 GPUs. One GPU slice is the same as 1/7th of an A100 GPU. The default is 0. You can specify the type of GPU as well: [Type]:<number> (e.g. A100:2). You can also just request a number of GPUs as <number>.**The interactive partition only allows you to request max 1 GPU (slice) though.**
+- Number of GPUs. Depending on the partition you have requested, you might get a different GPU.
```
w-lampaert, line 183: This is different now, as each GPU type now has its own partition. So you always get the GPU type of the partition.

source/leuven/services/openondemand.rst
```diff
@@ -141,47 +149,68 @@ the folder that it is associated with.
...
+  The default is 0.
+  The acquired GPU will be the same as the type specified in the partition (e.g. a NVidia H100 for ``gpu_h100`` on wICE).
+  For wICE, you can also request a GPU from the ``interactive`` partition.
+  One GPU here is a virtual GPU slice of the available A100 GPUs.
+  One GPU slice is the same as 1/7th of an A100 GPU.
+  You can specify the type of GPU as well: ``[Type]:<number>`` (e.g. ``A100:2``).
```
w-lampaert, line 189: This option has become somewhat redundant, as there is indeed only one GPU type per partition. Perhaps we can just remove it from the explanation?

source/leuven/services/openondemand.rst
```diff
@@ -191,136 +220,252 @@ of using any of the other partitions. There are however some limitations on the
...
-:ref:`R <create_r_conda_env>` conda environment. The second step consists out of effectively
-creating the kernel in your ``$VSC_HOME/.local`` folder, as Jupyter will look for kernels in this location. The following commands should be excecuted from a shell, 
-and only need to be done once for the set-up of each new kernel. This starts with activating your conda
-environment::
+To create any other kernel, first create a  :ref:`Python <create_python_conda_env>` or
+:ref:`R <create_r_conda_env>` Conda environment. The second step consists out of effectively
+creating the kernel in your ``$VSC_HOME/.local`` folder, as Jupyter will look for kernels in this location.
+The following commands should be excecuted from a shell (e.g. using 'Login Server Shell Access'), and only need
+to be done once for the set-up of each new kernel.
+If you already have an existing Python kernel, but your JupyterLab session freeses/craches when choosing your
```
w-lampaert, line 266: 'freezes'

source/leuven/services/openondemand.rst
```diff
@@ -191,136 +220,252 @@ of using any of the other partitions. There are however some limitations on the
...
+
+To launch MATLAB via OnDemand, you need to additionally specify your desired version of the software
+from the drop-down menu on the resource form.
+Given that our current MATLAB installations automatically detect GPU devices and CUDA libraries,
+you may also request GPU(s) as resources, if needed.
+
+Once you launch the session, a remote `noVNC`_ desktop will start on a compute node.
+Before clicking on the 'Launch MATLAB' button, one may still change the compression level and
+the image quality of the transferred noVNC frames.
+E.g. opting for the lowest compression level and highest image quality can give you a crisp VNC desktop.
```
w-lampaert, line 439: As we will have more and more VNC apps, it might be good to move the resolution part to the general resource description above.

source/leuven/services/openondemand.rst
```diff
@@ -191,136 +220,252 @@ of using any of the other partitions. There are however some limitations on the
...
+Once the session starts, the selected MATLAB module will be loaded, and eventually the MATLAB GUI
+will pop up (after waiting for few seconds).
+
+While your session is running, you can provide a view-only access to other VSC users.
+For that, click on the 'View Only (Share-able Link)' button to copy the URL into your clipboard,
```
w-lampaert, line 444: Same for the view-only. Perhaps move this to a more general part of the docs

source/leuven/services/openondemand.rst
```diff
@@ -191,136 +220,252 @@ of using any of the other partitions. There are however some limitations on the
...
+
+For visualization purposes, you may use the `ParaView app <https://www.paraview.org/>`_.
+Similar to the :ref:`MATLAB app <ood_matlab_app>`, ParaView also runs inside a `noVNC`_
+desktop as a compute job. An additional similarity to the :ref:`MATLAB app <ood_matlab_app>` is the choice for
+controlling the desktop resolution.
 
 **Remarks:**
 
-- Running lines of code is 'ctrl+enter' for R.
+- Currently, using GPUs in ParaVies is not supported yet, and just the CPU-only modules are offered.
```
w-lampaert, line 464: 'ParaView'

source/leuven/services/openondemand.rst
```diff
@@ -191,136 +220,252 @@ of using any of the other partitions. There are however some limitations on the
...
 
-For Python you will need the ``ipykernel`` package installed in your conda environment. Then you create the kernel as follows::
+For Python you will need the ``ipykernel`` package installed in your Conda environment::
 
-      python -m ipykernel install --user --name '<env_name>' --display-name '<kernel_name>'
+      conda install ipykernel
+
+Then you create the kernel as follows::
+
+      python -m ipykernel install --user --env PYTHONPATH "" --name '<env_name>' --display-name '<kernel_name>'
```
jooghe, line 279: I tried with an older kernel (Python 3.6.8, ipython 7.6.1,ipykernel 5.1.1) and got an error: 
 ipython-kernel-install: error: unrecognized arguments: --env PYTHONPATH
with a newer kernel (Python 3.8.16, ipython 8.8.0, ipykernel 6.19.2) re-registering worked and solved the issue. 

do we need  a minimum version to fix this?

source/leuven/services/openondemand.rst
moravveji: done

source/leuven/services/openondemand.rst
moravveji: Do you want me to remove the rest of the text after the first sentence?
I think the rest of the information are still valid, no?

source/leuven/services/openondemand.rst
moravveji: done

source/leuven/services/openondemand.rst
moravveji: done

source/leuven/services/openondemand.rst
moravveji: perhaps, yes.

source/leuven/services/openondemand.rst
```diff
@@ -191,136 +226,244 @@ of using any of the other partitions. There are however some limitations on the
...
+You should :ref:`install Miniconda <install_miniconda_python>` if you have not installed it yet.
 
-To create any other kernels, first create a  :ref:`Python <create_python_conda_env>` or 
-:ref:`R <create_r_conda_env>` conda environment. The second step consists out of effectively
-creating the kernel in your ``$VSC_HOME/.local`` folder, as Jupyter will look for kernels in this location. The following commands should be excecuted from a shell, 
-and only need to be done once for the set-up of each new kernel. This starts with activating your conda
-environment::
+To create any other kernel, first create a  :ref:`Python <create_python_conda_env>` or
+:ref:`R <create_r_conda_env>` Conda environment. The second step consists out of effectively
+creating the kernel in your ``$VSC_HOME/.local`` folder, as Jupyter will look for kernels in this location.
```
MaximeVdB, line 269: I know you only changed the linebreaks here, but maybe it would be good to fix this part. Because it's not always ``$VSC_HOME/.local``, isn't it rather ``$XDG_DATA_HOME``?

source/leuven/services/openondemand.rst
```diff
@@ -191,136 +226,244 @@ of using any of the other partitions. There are however some limitations on the
...
+
+      python -m ipykernel install --user --env PYTHONPATH "" --name '<env_name>' --display-name '<kernel_name>'
 
 For R, you need both the ``jupyter_client`` and the ``irkernel`` package installed. With the following command you can create the kernel::
 
       Rscript -e 'IRkernel::installspec(name="<env_name>", displayname="<kernel_name>")'
 
-Once the kernel is created, you will see it in the 'Launcher' menu. You can now start working in your own customized environment.
+Once the kernel is (re)created, you will see it in the 'Launcher' menu. You can now start working in your own customized environment.
+Note that all user kernels are stored by default in ``${VSC_HOME}/.local/share/jupyter/kernels``, but we strongly advice you
```
MaximeVdB, line 291: (same remark here as above, about ``$XDG_DATA_HOME``)

source/leuven/services/openondemand.rst
```diff
@@ -191,136 +226,244 @@ of using any of the other partitions. There are however some limitations on the
...
+
+Furthermore, you may optionally choose to load ``SciPy-bundle`` (for widely-used packages like ``scipy``,
+``numpy``, ``pandas`` and more) and/or ``matplotlib`` in your environment from the same 'Toolchain year'.
+
+Once you launch a JupyterLab session, a default kernel called ``Python 3 (ipykernel)`` is already available in your session.
+This kernel, in addition to the Python standard library, would enable using extra packages from
+``SciPy-bundle`` and/or ``matplotlib``, if you already selected them.
+
+If the "standard" environment explained above does not provide all packages that you need, then
+it is recommended to manage your custom-made R or Python environments using Conda.
```
MaximeVdB, line 264:
```suggestion
it is recommended to manage your custom-made R or Python environments using Conda or Python virtual environments.
```

source/leuven/services/openondemand.rst
```diff
@@ -191,136 +226,244 @@ of using any of the other partitions. There are however some limitations on the
...
 **Remarks:**
 
-- The start location is `$VSC_DATA`
-- At the moment, we do not support installing extensions in Jupyter Lab
+- The top-level notebook directory is ``$VSC_DATA``.
+- At the moment, we do not support installing extensions in JupyterLab.
+- One may create a customized kernel starting from a Python virtual environment, using one of Python modules.
+  This approach has interesting advantages, such as isolating a specific work environment, and also
+  installing packages into your data directory using ``pip install --prefix=${VSC_DATA}/<project-folder> <package-names>``.
+  If you are interested in this approach, contact your local HPC support for step-by-step instructions.
```
MaximeVdB, line 303: (So this is the `Remark` I was referring to in my comments above and which I'd like to see converted in a separate paragraph with more info. IMO this is something that should also be illustrated here, referring to HPC support is not sufficient.)

source/leuven/services/openondemand.rst
```diff
@@ -191,136 +226,244 @@ of using any of the other partitions. There are however some limitations on the
...
 Tensorboard
 -----------
 
-Tensorboard is an app that allows you to visualize and measure different aspects of your machine learning workflow. Have a look at the `official guidelines <https://www.tensorflow.org/tensorboard/get_started>`_ for more detailed information. 
+Tensorboard is an app that allows you to visualize and measure different aspects of your machine learning workflow.
+Have a look at the `official guidelines <https://www.tensorflow.org/tensorboard/get_started>`_ for more detailed information.
+
+The Tensorboard interactive session requires you to specify a project (or log) directory in your submission options.
+This is a relative directory starting from your ``$VSC_DATA``.
+It is not possible to navigate to the correct folder from within the app.
```
MaximeVdB, line 345: I don't really understand this sentence (in particular what 'correct folder' means here).

source/leuven/services/openondemand.rst
```diff
@@ -191,136 +226,244 @@ of using any of the other partitions. There are however some limitations on the
...
+
+Once you launch the session, a remote `noVNC`_ desktop will start on a compute node.
+Once the session starts, the selected MATLAB module will be loaded, and eventually the MATLAB GUI
+will pop up (after waiting for few seconds).
+
+.. warning::
+
+   As the end-user, you are responsible for all consequences of sharing your MATLAB session with other
+   VSC users.
+   So, think twice before sharing your sensitive data, sources and information by all means.
```
MaximeVdB, line 450: To me this comes a bit 'out of the blue'. What is meant here with 'sharing a MATLAB session with other users'? I suppose that would also help understand what would be MATLAB-specific about this concern.

MaximeVdB (changes requested):
Thanks Ehsan for taking care of this -- here are some comments (will probably need another round afterwards though)

source/leuven/services/openondemand.rst
moravveji: Correct. I moved the warning to where it belongs more naturally, i.e. just below the explanation for the 'View Only' button.

source/leuven/services/openondemand.rst
moravveji: Actually, this is a what we have on the official docs, since this page was created/approved. I also do not fully get it.
But, let me ping @w-lampaert for his idea on this.

source/leuven/services/openondemand.rst
```diff
@@ -191,136 +226,244 @@ of using any of the other partitions. There are however some limitations on the
...
-memory nodes), you can always request nodes from another partition. Remember however that these interactive apps are not meant for running full jobs. If you
-indeed need multiple nodes or full GPUs to test your code/program, go ahead and request the resources for your interactive app. In the case that you have passed
-the testing phase and you want to start conducting experiments, we recommend that you make the switch to batch jobs instead, as they will not require
-your presence to start your code.
+If for some reason some of these limitations are too strict for you, or you need resources that are not available on
+the interactive nodes (e.g. a full GPU, big memory nodes), you can always request nodes from another partition.
+Remember however that these interactive apps are not meant for running full jobs.
+If you indeed need multiple nodes or full GPUs to test your code/program, go ahead and request the resources for
+your interactive app from a more suited partition.
+In the case that you have passed the testing phase, and you want to start conducting experiments;
```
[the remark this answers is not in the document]
moravveji: leftover from the past. changed `;` with `,`

source/leuven/services/openondemand.rst
```diff
@@ -191,136 +226,244 @@ of using any of the other partitions. There are however some limitations on the
...
+JupyterLab
 -----------
 
-With this app you can use and create Jupyter Notebooks. This can be handy both for R and Python coding. There are two kernels already available, being a Python and a
-R kernel. The Python and R versions that are used for this, are the versions located in ``/usr/bin``. While you can use these to do some testing, it is not recommended 
-to work with these. It is better to work with conda environments in this case. You should :ref:`install miniconda <install_miniconda_python>` if you have not installed it yet.
-When you do not have any conda environments and their associated kernels, both the Python and R installation will default to the ``~/miniconda3/bin/...`` installation.
+With this app you can create or (re)run Jupyter Notebooks. This can be handy both for R and Python coding.
+In the app resource form, besides the normal choices (:ref:`listed above <interactive-apps>`), you can also choose
+a 'Toolchain year' such as '2023a' from a drop-down menu.
```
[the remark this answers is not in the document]
moravveji: I have made the change. I have to also remember to change the field title on the form from 'Toolchain year' to 'Toolchain version' later.

source/leuven/services/openondemand.rst
moravveji: Is it better now?

source/leuven/services/openondemand.rst
moravveji: Sure. Let me know what you think here.

source/leuven/services/openondemand.rst
```diff
@@ -191,136 +226,244 @@ of using any of the other partitions. There are however some limitations on the
...
+Based on that choice, a correct JupyterLab module will be loaded together with its dependencies (specifically Python).
+To reproduce your numerical results at any time in the future, make sure you stick to choosing the same 'Toolchain year'.
```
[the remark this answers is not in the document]
moravveji: I elaborated on how to create and use kernels from python venv. As you see above, it might be involved for some users. So, the question would be whether we want to support it or not. Let's take this internal discussion outside this PR.

moravveji (commented):
Including suggestions from Maxime

source/leuven/services/openondemand.rst
moravveji: Sure, but I did it slightly differently now.

source/leuven/services/openondemand.rst
moravveji: Indeed. I turned this one into a sub-section, so that it stands out.

source/leuven/services/openondemand.rst
moravveji: I checked few conda environments, and I updated the text accordingly:

"The minimum supported version for Python is 3.7 and for ``ipykernel`` package is 6.19.2."

source/leuven/services/openondemand.rst
w-lampaert: The part about the interactive partition is rather relevant for ondemand though, as this is the preferred partition where users submit their interactive sessions to. So at least mentioning that you can get a GPU on the interactive partition seems important. The whole part about the type of GPU, yes I would remove that.

moravveji: The updated text already reflects your idea. But, you might want to check it again.

source/leuven/services/openondemand.rst
```diff
@@ -191,136 +241,297 @@ of using any of the other partitions. There are however some limitations on the
...
+Eventhough this approach can have advantages, but we do not recommend it, due to the fact that Python
+virtual environments are architecture specific.
+Hence, on our current Tier-2 machines, this approach needs to be handled with care.
+If you are interested in this approach, follow these steps to create and use your kernel:
+
+- Pick a specific Python from a specific 'Toolchain version', e.g. ``Python/3.11.3-GCCcore-12.3.0``
+  from ``2023a``
+- Choose a specific architecture, e.g. Sapphire Rapids nodes on wICE
+- Start an :ref:`Interactive Shell<interactive_shell>` on the targeted architecture,
+  and execute the following:
```
MaximeVdB, line 332: I think it is best to have 'basic' and 'advanced' scenarios:
- What is currently written here would correspond to the 'advanced' one, where the user anticipates that architecture-dependent installations will be needed. This is normally only needed under certain circumstances (e.g. when installing some packages requires compiling extensions from source).
- So probably best to start with a simpler 'basic' scenario (before showing the more advanced case) where the architecture dependence is not explicitly considered, e.g.:
```bash
     python -m venv ${VSC_DATA}/venv_science_2023a
     (etcetera)
```

source/leuven/services/openondemand.rst
```diff
@@ -191,136 +241,297 @@ of using any of the other partitions. There are however some limitations on the
...
 
-For R, you need both the ``jupyter_client`` and the ``irkernel`` package installed. With the following command you can create the kernel::
+
+Custom virtual environments for Python
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+One may create a customized kernel starting from a Python virtual environment, using one of Python modules.
+Eventhough this approach can have advantages, but we do not recommend it, due to the fact that Python
+virtual environments are architecture specific.
+Hence, on our current Tier-2 machines, this approach needs to be handled with care.
```
MaximeVdB, line 325: `due to the fact that Python virtual environments are architecture specific.`  -> this is not the case. Just as with Conda environments, it mostly depends on how you install the packages. If you're installing precompiled packages (e.g. from PyPI repos via `pip install ...` or from Conda repos via `conda install ...`), then these will have been compiled in such a way that compiled parts (if any) can run on any (x86) architecture (except perhaps very old ones).

When using Python installations from a module, then of course these Python modules are replicated for the different available architectures (Skylake, Cascadelake, ...). In practice, however, it's no problem to create a virtual environment with some `Python/3.X.Y-...` module on one architecture (say, Skylake) and then use this venv on another architecture (say, Genoa). At least that's what I've experienced so far. It's of course best to load `Python/3.X.Y-...` module for the architecture at hand (and not that of another architecture), but that's already how the module-based-JupyterLab-app is operating.

Also: while I wouldn't say we should explicitly recommend virtual environments (nor Conda environments, for that matter), I don't think we should discourage it either.

MaximeVdB (changes requested):

source/leuven/services/openondemand.rst
MaximeVdB: (I think this is now further discussed in a new thread)

source/leuven/services/openondemand.rst
```diff
@@ -191,136 +241,297 @@ of using any of the other partitions. There are however some limitations on the
...
+Custom Conda environments
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you have not installed Conda in your account yet, please refer to the
+:ref:`install Miniconda <install_miniconda_python>` page.
+
+To create a custom kernel, first create a  :ref:`Python <create_python_conda_env>` or
+:ref:`R <create_r_conda_env>` Conda environment. The second step consists of effectively
+creating the kernel in your ``$VSC_HOME/.local/share`` folder (which is a defalt value for the
+``$XDG_DATA_HOME`` environment variable), as Jupyter will look for kernels in this location.
```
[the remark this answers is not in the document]
moravveji: I added it as a remark at the bottom of the same section.

source/leuven/services/openondemand.rst
```diff
@@ -191,136 +241,297 @@ of using any of the other partitions. There are however some limitations on the
...
-For Python you will need the ``ipykernel`` package installed in your conda environment. Then you create the kernel as follows::
+For Python you will need the ``ipykernel`` package installed in your Conda environment.
+The minimum supported version for Python is 3.7 and for ``ipykernel`` package is 6.19.2::
+
+      conda install ipykernel
+
+Then you create the kernel as follows::
 
-      python -m ipykernel install --user --name '<env_name>' --display-name '<kernel_name>'
+      python -m ipykernel install --user --env PYTHONPATH "" --name '<env_name>' --display-name '<kernel_name>'
```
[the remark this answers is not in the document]
moravveji: see above.

source/leuven/services/openondemand.rst
```diff
@@ -191,136 +241,297 @@ of using any of the other partitions. There are however some limitations on the
...
+For R, you need both the ``jupyter_client`` and the ``irkernel`` packages installed.
+With the following command you can create the kernel::
 
       Rscript -e 'IRkernel::installspec(name="<env_name>", displayname="<kernel_name>")'
 
-Once the kernel is created, you will see it in the 'Launcher' menu. You can now start working in your own customized environment.
+Once the kernel is created, you will see it in the 'Launcher' menu.
+You can now start working in your own customized environment.
+Note that all user kernels are stored by default in ``${VSC_HOME}/.local/share/jupyter/kernels``
+(where ``${VSC_HOME}/.local/share`` is the default value for ``$XDG_DATA_HOME``);
```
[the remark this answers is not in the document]
moravveji: Agreed. I have created a 'Note' block at the bottom of the JupyterLab section, to elaborate a bit about `XDG_DATA_HOME`.

source/leuven/services/openondemand.rst
```diff
@@ -191,136 +241,297 @@ of using any of the other partitions. There are however some limitations on the
...
+
+     TOOLCHAIN='2023a'
+     DIR_VENV=${VSC_DATA}/${VSC_OS_LOCAL}/${VSC_ARCH_LOCAL}${VSC_ARCH_SUFFIX}/${TOOLCHAIN}
+     mkdir -p ${DIR_VENV}
+     module use /apps/leuven/${VSC_OS_LOCAL}/${VSC_ARCH_LOCAL}${VSC_ARCH_SUFFIX}/${TOOLCHAIN}/modules/all
+     module load Python/3.11.3-GCCcore-12.3.0
+     python -m venv ${DIR_VENV}
+     source ${DIR_VENV}/bin/activate
+     pip install --prefix=${DIR_VENV} ipykernel <additional packages>
+     # note that below, --env is not needed
```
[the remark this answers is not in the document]
moravveji: done

source/leuven/services/openondemand.rst
```diff
@@ -191,136 +241,297 @@ of using any of the other partitions. There are however some limitations on the
...
+     python -m ipykernel install --user --name '<env_name>' --display-name '<kernel_name>'
+     # you need this below
+     echo ${DIR_VENV}
+
+- On the JupyterLab form, choose a partition that matches your choice of architecture;
+  also, pick the same 'Toolchain version' as above.
+  In the 'Pre-run scriptlet' box, insert ``source <path/pointing/to/DIR_VENV>/bin/activate``.
```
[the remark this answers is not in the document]
moravveji: Correct. I took that line out.

source/leuven/services/openondemand.rst
```diff
@@ -191,136 +241,297 @@ of using any of the other partitions. There are however some limitations on the
...
+- Choose a specific architecture, e.g. Sapphire Rapids nodes on wICE
+- Start an :ref:`Interactive Shell<interactive_shell>` on the targeted architecture,
+  and execute the following:
+
+  .. code-block :: bash
+
+     TOOLCHAIN='2023a'
+     DIR_VENV=${VSC_DATA}/${VSC_OS_LOCAL}/${VSC_ARCH_LOCAL}${VSC_ARCH_SUFFIX}/${TOOLCHAIN}
+     mkdir -p ${DIR_VENV}
+     module use /apps/leuven/${VSC_OS_LOCAL}/${VSC_ARCH_LOCAL}${VSC_ARCH_SUFFIX}/${TOOLCHAIN}/modules/all
```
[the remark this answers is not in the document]
moravveji: This is unfortunately needed when using interactive shell apps, because cluster modules are not automatically loaded. See https://gitea.icts.kuleuven.be/coz/ood/issues/13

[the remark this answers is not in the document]
moravveji: I added a line of comment above it, to explain when the `module use ...` is needed.

source/leuven/services/openondemand.rst
```diff
@@ -191,136 +241,297 @@ of using any of the other partitions. There are however some limitations on the
...
+- Pick a specific Python from a specific 'Toolchain version', e.g. ``Python/3.11.3-GCCcore-12.3.0``
+  from ``2023a``
+- Choose a specific architecture, e.g. Sapphire Rapids nodes on wICE
+- Start an :ref:`Interactive Shell<interactive_shell>` on the targeted architecture,
+  and execute the following:
+
+  .. code-block :: bash
+
+     TOOLCHAIN='2023a'
+     DIR_VENV=${VSC_DATA}/${VSC_OS_LOCAL}/${VSC_ARCH_LOCAL}${VSC_ARCH_SUFFIX}/${TOOLCHAIN}
```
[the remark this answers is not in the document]
moravveji: yes it is.

source/leuven/services/openondemand.rst
moravveji: The text in the JupyterLab secion has already grown, and I prefer to keep it as least complex as possible. Splitting a sub-section into two scenarios just adds to the length of the text and its complexity, which I strongly like to avoid here.
Users who want to create a kernel out of Python venv are already advanced users, so they shall follow only one set of instructions. This is preferable on the support-side.
For that reason, I prefer to keep only one scenario here.

source/leuven/services/openondemand.rst
moravveji: I do not think so. Let's work it out by a way of example.
I have created a kernel from a venv, starting from `Python/3.11.3-GCCcore-12.3.0` on the wICE `interactive` partition. As a result of that, the first lines of my `kernel.json` read:
```
{
 "argv": [
  "/vsc-hard-mounts/leuven-data/307/vsc30745/tests/ondemand/jupyterlab/test_venv_kernel/bin/python",
...
```
when I look up the `python` exactuable pointed at above, I see that it is a softlink pointing at the Python executable on IceLake nodes:
```
ls -l /vsc-hard-mounts/leuven-data/307/vsc30745/tests/ondemand/jupyterlab/test_venv_kernel/bin/python
lrwxrwxrwx 1 vsc30745 vsc30745 82 Jun 14 10:40 /vsc-hard-mounts/leuven-data/307/vsc30745/tests/ondemand/jupyterlab/test_venv_kernel/bin/python -> /apps/leuven/rocky8/icelake/2023a/software/Python/3.11.3-GCCcore-12.3.0/bin/python
```
Now, if I start a job on any architecture other than IceLake nodes, my kernel wil l freeze/crach (I tested this and I saw it).

The reason for the issue is not the extra packages that we install with `pip` into our venv, but is the non-portability of the architecture-specific binaries.

w-lampaert: Perhaps this can be explained more correctly then. I would omit the part that says that it is architecture-specific, but rather say something like:

'If you want to use Python venvs, be aware that you can only use that venv for the specific architecture you created it on, as the kernel will link to the specific Python module/version it was created with. This means that on other architectures, that module won't be available, which will result in the crashing of your kernel.'

source/leuven/services/openondemand.rst
```diff
@@ -191,136 +241,308 @@ of using any of the other partitions. There are however some limitations on the
...
+the interactive nodes (e.g. a full GPU, big memory nodes), you can always request nodes from another partition.
+Remember however that these interactive apps are not meant for running full jobs.
+If you indeed need multiple nodes or full GPUs to test your code/program, go ahead and request the resources for
+your interactive app from a more suitable partition.
+In the case that you have passed the testing phase, and you want to start conducting experiments,
+we recommend that you make the switch to batch jobs instead, as they will not require your presence to start your code.
 
 .. _interactive_shell:
 
 Interactive shell
```
w-lampaert, line 254: Perhaps we should shortly explain the issue with the module loading here as well (https://gitea.icts.kuleuven.be/coz/ood/issues/13)

source/leuven/services/openondemand.rst
w-lampaert: The input data for tensorboard are always log files from a ML run. These get stored in a log directory by default. The Tensorboard app form has a project/log directory variable to specify, which should be the path to the log directory with the data you want to analyze. If you do not specify the correct directory here, the Tensorboard app will launch, but immediately give an error. Once within the session, you cannot change your working directory anymore. You can only cancel the job and start a new one with the correct directory specified under the project/log dir variable.

source/leuven/services/openondemand.rst
jooghe: To make it very clear put it in a "note: " ?

source/leuven/services/openondemand.rst
MaximeVdB: Ehsan: could you describe your example in more detail? Because I myself cannot reproduce it (i.e. so far I haven't had problems of that sort), perhaps my own test is too simple:

Just now, I created a venv based on a Python/3.9.5-... module from `icelake/2021a` on a wICE interactive node in which I installed NumPy and then created the kernel for it. I then used OOD in test to launch JupyterLab (with 2021a toolchain version; no scriptlet) on the test batch partition, so with skylake-6132 CPUs. I could just select my kernel for a new notebook and e.g. calculate some dot products with NumPy. And just to emphasize: in this situation, the Python interpreter used is the skylake-6132-installed one (which is indeed a requirement for this to work).

moravveji: I admit that it is possible to create venv on one arch and use it on another arch, as I verified on both sandbox skylake nodes and wice sapphire rapids nodes.
The commit 9808326 shall reflect the changes to the text under the venv section.

MaximeVdB (changes requested):

source/leuven/services/openondemand.rst
```diff
@@ -191,136 +241,324 @@ of using any of the other partitions. There are however some limitations on the
...
+
+      python -m ipykernel install --user --env PYTHONPATH "" --name '<env_name>' --display-name '<kernel_name>'
 
-      python -m ipykernel install --user --name '<env_name>' --display-name '<kernel_name>'
 
-For R, you need both the ``jupyter_client`` and the ``irkernel`` package installed. With the following command you can create the kernel::
+Custom virtual environments for Python
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+One may create a customized kernel starting from a Python virtual environment, using one of Python modules.
```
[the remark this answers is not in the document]
moravveji: done

source/leuven/services/openondemand.rst
```diff
@@ -191,136 +241,324 @@ of using any of the other partitions. There are however some limitations on the
...
+.. note::
+
+   User kernels are stored by default in ``${VSC_HOME}/.local/share/jupyter/kernels``.
+   To override this and store your kernel specifications in a non-default location,
+   you may drop the following line in your ``${VSC_HOME}/.bashrc``::
+
+      export XDG_DATA_HOME=${VSC_DATA}/.local/share
+
+   If you have created multiple kernels (for Python and R), you find the corresponding
+   directories under ``${XDG_DATA_HOME}/jupyter/kernels``.
```
MaximeVdB, line 397: I found this to be a bit unclear and I think this would improve it:
```suggestion
   When the ``${XDG_DATA_HOME}`` variable is set, subsequent kernel installations
   (for both Python and R) will reside in ``${XDG_DATA_HOME}/jupyter/kernels``.
```

MaximeVdB (changes requested):
(now that the larger changes got dealt with, I could now focus on hopefully final polishings and restructurings ;)

source/leuven/services/openondemand.rst
```diff
@@ -191,136 +241,324 @@ of using any of the other partitions. There are however some limitations on the
...
-:ref:`R <create_r_conda_env>` conda environment. The second step consists out of effectively
-creating the kernel in your ``$VSC_HOME/.local`` folder, as Jupyter will look for kernels in this location. The following commands should be excecuted from a shell, 
-and only need to be done once for the set-up of each new kernel. This starts with activating your conda
-environment::
+   If you use JupyterLab in this way, remember to be consistent in your choice of toolchain version
+   as this e.g. determines the version of Python that will be used.
+
+If these modules does not provide all Python packages that you need, then you can locally install
+these extra packages, followed by installing the corresponding Jupyter kernel  (see
+"Custom virtual environments for Python" below).
```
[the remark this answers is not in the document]
moravveji: done in 8f94b21.

source/leuven/services/openondemand.rst
```diff
@@ -191,136 +241,324 @@ of using any of the other partitions. There are however some limitations on the
...
+
+Conda environments for Python
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you have not installed Conda in your account yet, please refer to the
+:ref:`install Miniconda <install_miniconda_python>` page.
+
+To create a custom kernel, first create a  :ref:`Python <create_python_conda_env>` or
```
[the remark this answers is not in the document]
moravveji: done in 07f2f6d

source/leuven/services/openondemand.rst
```diff
@@ -191,136 +241,324 @@ of using any of the other partitions. There are however some limitations on the
...
+For more general information, please refer to the `official JupyterLab documentation`_.
 
 **Remarks:**
 
-- The start location is `$VSC_DATA`
-- At the moment, we do not support installing extensions in Jupyter Lab
+- The top-level notebook directory is by default ``$VSC_DATA``.
+- At the moment, we do not support installing extensions in JupyterLab.
+
+.. note::
```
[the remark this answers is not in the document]
moravveji: In 1c9bf78, I moved up the remarks to the beginning of the JupyterLab section, before the "Pure Python environments".

However, the scope of the 'Note' box applies to all Python/R Conda/venv kernels. 
The problem with (our) restructured text is that the subsectioning is not fully clear, eventhough we have consistently used correct underline characters for the (sub)section titles. As a result, the 'Note' box appears below the R Conda section, but in practice it shall belong to all kernel-driven envs.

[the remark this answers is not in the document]
MaximeVdB: Well, you could still move the Note up to the beginning of the section, right? Then that's also where it will appear. But on second thought it is probably best to not move that Note to the top (only the remarks, as you just did), since it's too detailed to be placed at the beginning.

How about this:
First it would indeed be great if the (sub, subsub, ...)section titles on this page would get numbered. I don't think that really has to do with how the titles are underlined, but how the page's [TOC tree](https://documatt.com/restructuredtext-reference/element/toctree.html) is set up. Could you look into that?
Then it would just be a matter of adding another `Remarks` subsection after the R Conda subsection to include additional notes (such as this one) and the numbering will then make clear that this applies to the whole JupyterLab section.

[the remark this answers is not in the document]
moravveji: I tried out enumerating the (sub)sections, and it did not turn out a success, because:
- the 'Note' box did still sit below the same subsection for 'Conda environment for R', and it did not look like belonging to the whole JupyterLab section
- the section enumeration does not impose indenting the subsections, or making them recognizable. So, the ambiguity stayed.

Instead, I created a new sub(sub?)section under the JupyterLab section, and move the 'Note' block up inside this one. You may see it in 1e7a459.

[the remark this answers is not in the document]
moravveji: @MaximeVdB what about the changes in c24eff6? The sections are only numbered, without (a long) table of content.

source/leuven/services/openondemand.rst
```diff
@@ -191,136 +243,333 @@ of using any of the other partitions. There are however some limitations on the
...
+
+To launch MATLAB via OnDemand, you need to additionally specify your desired version of the software
+from the drop-down menu on the resource form.
+Given that our current MATLAB installations automatically detect GPU devices and CUDA libraries,
+you may also request GPU(s) as resources, if needed.
+
+Once you launch the session, a remote `noVNC`_ desktop will start on a compute node.
+Once the session starts, the selected MATLAB module will be loaded, and eventually the MATLAB GUI
+will pop up (after waiting for few seconds).
+
```
jooghe, line 558: Best to add here an additional remark, that Matlab will only be usable by users that are in a license group for Matlab.
This remark can be site specific, as for us this is group lli_matlab but could be something else for Gent, Antwerp or VUB.

source/leuven/services/openondemand.rst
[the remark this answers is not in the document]
MaximeVdB: Yes that looks pretty good (the only downside being that the page title itself also gets numbered; but it looks like there is not an easy way to avoid that, so for the time being this is OK for me).

MaximeVdB (approved):

w-lampaert (approved):

source/leuven/services/openondemand.rst
moravveji: Given that this page is already site-specific, I just added a 'Note' block about the license. Please see 1e67e1a.

jooghe (approved):

PR #426, opened 2024-08-19
OOD RStudio & R package management documentation update

WouterVanAssche:


MaximeVdB (changes requested):

source/leuven/services/openondemand.rst
```diff
@@ -419,9 +419,16 @@ For more general information, please refer to the `official JupyterLab documenta
...
-You will be running RStudio with R version 4.2.1.
-For more information on how to use RStudio, check out the `RStudio official documentation`_.
+This interactive app allows you to run an RStudio session on the cluster. 
+This R session makes use of the R/4.2.2-foss-2022b module. 
+Additionally, the R-bundle-CRAN and R-bundle-Bioconductor modules can be loaded 
+on top of the base R to provide easy access to hundreds of preinstalled packages.
+
+You can use your own personal R libraries with Studio, see :ref:`R package management<r_package_management_standard_lib>`. 
+While the use of R projects is also available within RStudio, loading such project may affect your module library paths as well as your personal R library path.
+We therefore recommend to only use R projects with renv, ensuring a completely isolated library for your project.
```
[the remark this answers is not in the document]
WouterVanAssche: The use of R projects is actually best practice and we really shouldn't encourage against it. The whole point of a project is to create a separate work environment for your project. From my previous testing, projects that use renv work well with RStudio. R projects are also not something every R user will be using, those that do use them are typically more experienced users. I think it is therefore better to inform them on the correct way of using R projects.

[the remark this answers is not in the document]
MaximeVdB: (In the meantime we agreed that this should be reformulated to make clear that using "RStudio Projects" (not "R Projects") is fine provided that the user ticks the "use renv for this project" option.)

source/leuven/services/openondemand.rst
```diff
@@ -419,9 +419,16 @@ For more general information, please refer to the `official JupyterLab documenta
...
+
+For more information on how to use RStudio, check out the `official documentation <https://docs.rstudio.com/>`__.
 
 The use is very similar to regular RStudio.
 It is recommended to install packages in a folder on your ``$VSC_DATA`` instead of the default location though,
```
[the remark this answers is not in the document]
MaximeVdB: indeed, this is already touched upon on that other page and that page is already referenced here

source/leuven/services/openondemand.rst
```diff
@@ -419,14 +419,17 @@ For more general information, please refer to the `official JupyterLab documenta
...
+Additionally, the R-bundle-CRAN and R-bundle-Bioconductor modules can be loaded 
+on top of the base R module to provide easy access to hundreds of preinstalled packages.
 
-The use is very similar to regular RStudio.
-It is recommended to install packages in a folder on your ``$VSC_DATA`` instead of the default location though,
-to avoid clogging your ``$VSC_HOME``.
-You can do this by using the ``lib`` argument for both the ``install.packages`` and the ``library`` function.
+It is also possible to use locally installed R packages with RStudio, see :ref:`R package management<r_package_management_standard_lib>`. 
+While the use of RStudio projects is also available within RStudio, loading such project may affect your module library paths as well as your personal R library path.
+We therefore recommend to only use RStudio projects with `renv <https://rstudio.github.io/renv/articles/renv.html>`__, ensuring a completely isolated library for your project.
```
MaximeVdB, line 430: I think the following formulation will be easier to follow:
```suggestion
RStudio furthermore allows to create RStudio projects to manage your
R environments. When doing so, we recommend to select the
`renv <https://rstudio.github.io/renv/articles/renv.html>`_ option
to ensure a completely independent R environment. Without `renv`,
loading an RStudio project may lead to incomplete R library paths.
```

source/software/r_package_management.rst
```diff
@@ -6,51 +6,106 @@ R package management
...
+the recommended location to install software and e.g. `$VSC_DATA` should be used
+instead.
+
+Secondly, it should be kept in mind that R packages often include extensions written in
+compiled languages (e.g. C++ or Fortran) and that the centrally installed R modules are
+configured to compile these extensions with optimizations for the CPU architecture at hand.
+This means that such R packages cannot in general be used on different partitions than the
+one they were created on.
+
+Thirdly, R packages may also only work with certain versions of R and not with other version.
```
MaximeVdB, line 39:
```suggestion
Thirdly, R packages may also only work with certain versions of R and not with other versions.
```

source/leuven/services/openondemand.rst
```diff
@@ -419,14 +419,17 @@ For more general information, please refer to the `official JupyterLab documenta
...
+
+For more information on how to use RStudio, check out the `official documentation <https://docs.rstudio.com/>`__.
```
MaximeVdB, line 432: https://docs.rstudio.com redirects to https://docs.posit.com and it looks like it would be better to link to a more RStudio-specific page on that website:
```suggestion
For more information on how to use RStudio, check out the `official documentation <https://docs.posit.co/ide/user/>`__.
```

MaximeVdB (changes requested):

MaximeVdB (approved):

source/leuven/services/openondemand.rst
```diff
@@ -419,14 +419,20 @@ For more general information, please refer to the `official JupyterLab documenta
...
+Additionally, the R-bundle-CRAN and R-bundle-Bioconductor modules can be loaded 
+on top of the base R module to provide easy access to hundreds of preinstalled packages.
 
-The use is very similar to regular RStudio.
-It is recommended to install packages in a folder on your ``$VSC_DATA`` instead of the default location though,
-to avoid clogging your ``$VSC_HOME``.
-You can do this by using the ``lib`` argument for both the ``install.packages`` and the ``library`` function.
+It is also possible to use locally installed R packages with RStudio, see :ref:`R package management<r_package_management_standard_lib>`. 
+RStudio furthermore allows to create RStudio projects to manage your
+R environments. When doing so, we recommend to select the
```
moravveji, line 430: `R` and everywhere else perhaps?
We also have to double check this within our team perhaps, and across other pages. 
E.g. Python is not typeset as `Python`; so maybe the same also applies to R (or `R` ;-) )

source/leuven/services/openondemand.rst
```diff
@@ -436,16 +442,15 @@ You can do this by using the ``lib`` argument for both the ``install.packages``
   You will also notice that you cannot use the same way of navigating after this.
   Another solution is to click the three dots on the right (...) and enter your path.
 - The 'Tools-Install packages' interface does not allow you to select any other path than the default in your ``$VSC_HOME``.
-  It is recommended to use the ``install.packages`` function instead.
+  It is recommended to use the ``install.packages()`` function instead.
 - RStudioServer will by default store the RStudio cache in ``$VSC_HOME/.local/share/rstudio``.
   This cache can get very large, and cause you to exceed the quota of your home directory.
-  To avoid this, you can redirect this cache to your data directory by setting ``$XDG_DATA_HOME``
-  variables in your ``~/.bashrc``.
+  To avoid this, you can redirect this cache to your scratch directory by setting the ``$XDG_DATA_HOME``
```
moravveji, line 448: In section 7.3.2, we are recommending the users to set `XDG_DATA_HOME` to somewhere in their data folder! Clearly, this will be confusing for the users.
Each of these two choices have pros and cons; hence, I recommend to make a decision within our team and fix this in this page coherently.

source/software/r_package_management.rst
```diff
@@ -6,51 +6,106 @@ R package management
...
 
-Most of the useful R packages can be installed separately. Some of those are 
-part of the centrally installed R modules. However, given the astounding number of 
-packages, it is not sustainable to install each and everyone of them system wide. 
-Fortunately, it is very easy for users to install R packages themselves.
-If you do encounter problems when doing so, do not hesitate to contact support.
+There exist thousands of R packages, available from online repositories like CRAN,
+Bioconductor or github. Depending on the R version, the more commonly used packages like `ggplot2`, `tidyverse` or `readr` 
+are either already included in the centrally installed R module or can be accessed by
+loading the R-bundle-CRAN and R-bundle-Bioconductor modules, e.g.:
```
moravveji, line 12: again  `R-bundle-CRAN` and `R-bundle-Bioconductor`

source/software/r_package_management.rst
```diff
@@ -6,51 +6,106 @@ R package management
...
+
+.. code:: r
+
+       $ module load R-bundle-Bioconductor/3.16-foss-2022b-R-4.2.2    
+
+It is possible, however, that these modules do not contain all R packages you need
```
moravveji, line 18: ... you need, or ...  (a comma is needed)

source/software/r_package_management.rst
```diff
@@ -6,51 +6,106 @@ R package management
...
+or that the package versions do not meet your requirements. In this case you will
+need to locally install those packages, as will be described below. Do not hesitate
+to contact support when encountering issues during these local installations.
```
moravveji, line 21: ... to contact your local support team ...

source/software/r_package_management.rst
```diff
@@ -6,51 +6,106 @@ R package management
...
 
 .. _r_package_management_standard_lib:
 
 Standard R package installation
 -------------------------------
 
-Setting up your own package repository for R is straightforward.
+Firstly, it is important to realize that R will use by default the `$VSC_HOME/R` path
```
moravveji, line 28: ... that R by default uses ...

source/software/r_package_management.rst
```diff
@@ -6,51 +6,106 @@ R package management
...
+to install new packages. Since home directories have limited quota, it is not
```
moravveji, line 29: ..., they are not ...

source/software/r_package_management.rst
```diff
@@ -6,51 +6,106 @@ R package management
...
+the recommended location to install software and e.g. `$VSC_DATA` should be used
+instead.
```
moravveji, line 31: ... to install software. Instead, we recommend to use `$VSC_DATA`.

source/software/r_package_management.rst
```diff
@@ -6,51 +6,106 @@ R package management
...
+
+Thirdly, R packages may also only work with certain versions of R and not with other versions.
+
+With these three considerations in mind, we recommend to use a directory structure which
+provides a unique path for each OS version, hardware architecture and R version.
+The example below creates such a structure for a Rocky8 OS, Icelake CPU and R version 4.2.2:
+
+.. code-block:: bash
+
+      $ mkdir -p ${VSC_DATA}/Rlibs/rocky8/icelake/R-4.2.2
```
moravveji, line 47: Similar to the next code block, I recommend using env vars for e.g. `rocky8` instead of hardcoding this.

source/software/r_package_management.rst
```diff
@@ -6,51 +6,106 @@ R package management
...
+
+Secondly, it should be kept in mind that R packages often include extensions written in
+compiled languages (e.g. C++ or Fortran) and that the centrally installed R modules are
+configured to compile these extensions with optimizations for the CPU architecture at hand.
+This means that such R packages cannot in general be used on different partitions than the
+one they were created on.
+
+Thirdly, R packages may also only work with certain versions of R and not with other versions.
+
+With these three considerations in mind, we recommend to use a directory structure which
```
moravveji, line 41: Your recommendation here also makes sense when the package installation happens as a job, and not on a login node! Of course using RStudio via OnDemand ensures it. using our typical `VSC_*` env vars ensures that even users who install their packages by mistake on the login node still get a unique library path for it, which would not be used when they submit a job.

source/software/r_package_management.rst
```diff
@@ -6,51 +6,106 @@ R package management
...
+      $ mkdir -p ${VSC_DATA}/Rlibs/rocky8/icelake/R-4.2.2
+
+The next step is to ensure such install locations are used by default in the R package installation process.
+This can be done by setting the `R_LIBS_USER` variable to in the `~/.Renviron` file as follows:
+
+.. code-block:: bash
+
+      $ echo 'R_LIBS_USER=${VSC_DATA}/Rlibs/${VSC_OS_LOCAL}/${VSC_ARCH_LOCAL}/R-${EBVERSIONR}' >> ~/.Renviron
+
+The `${VSC_OS_LOCAL}` and `${VSC_ARCH_LOCAL}` environment variables are predefined
```
moravveji, line 56: Again, your example makes sense when this has all happened as a compute job.

source/software/r_package_management.rst
```diff
@@ -6,51 +6,106 @@ R package management
...
+your packages in the appropriate R library folder.
 
-#. Load the appropriate R module, i.e., the one you want the R package
-   to be available for::
+.. note::
+
+  This `.Renviron` configuration will also work as expected in Open OnDemand apps
+  such as RStudio Server.
+
+The next step is to load the appropriate R module and launch R.
```
moravveji, line 69: ... R module, and run R.

source/software/r_package_management.rst
```diff
@@ -6,51 +6,106 @@ R package management
...
+
+.. code-block:: r
 
-      $ module load R/3.2.1-foss-2014a-x11-tcl
+      > install.packages("DEoptim")
 
-#. Start R and install the package (preferably in your $VSC_DATA directory)::
 
-      > install.packages("DEoptim", lib="/data/leuven/304/vsc30468/R/")
+If you are unsure whether R will install in the correct location, you can first list
```
moravveji, line 83: ... R will install your desired package in the correct ...

source/software/r_package_management.rst
```diff
@@ -6,51 +6,106 @@ R package management
...
+the known library locations by executing `.libPaths()`. The first location is the
+default one.
 
-#. Alternatively you can download the desired package::
+You can also specify your desired library path as an extra argument in the install command.
+This will take precedent over any defaults.
```
moravveji, line 88: precedence

source/software/r_package_management.rst
```diff
@@ -6,51 +6,106 @@ R package management
...
-   need to reinstall them for the other version.
-   
-Some R packages depend on libraries installed on the system.  In that case,
-you first have to load the modules for these libraries, and only then proceed
-to the R package installation.  For instance, if you would like to install
-the `gsl` R package, you would first have to load the module for the GSL
-library, .e.g., ::
-
-   $ module load GSL/2.5-GCC-6.4.0-2.28
+      $ R CMD INSTALL DEoptim_2.0-0.tar.gz  -l ${VSC_DATA}/Rlibs/rocky8/icelake/R-4.2.2
```
moravveji, line 105: For consistency, shall we also recommend using `VSC_*` env vars instead of hard-coded path?

moravveji (changes requested):

source/leuven/services/openondemand.rst
WouterVanAssche: I think this change won't improve readability. Back-quotes should be reserved for code snippets and perhaps explicit module names. But to back quote every single instance of `R` is a bit over kill. We should also start doing this for `RStudio`, `MATLAB`, `Python` etc. I personally think leaving software names in text as they are and only put back-quotes around code that should jump out so it is easier to find back and perhaps copy by the user.

source/leuven/services/openondemand.rst
WouterVanAssche: I am personally leaning towards advising scratch, especially for R users as the RStudio cache can get very large and may unknowingly occupy a large part of someone's data storage. We can definitely discuss this with the team.

WouterVanAssche: Maxime proposed a nice compromise. To stick with `XDG_DATA_HOME=$VSC_DATA/.local/share`. But also for RStudio specifically, so inform the users about some default settings to not by default save the RSession when RStudio is closed. This to avoid the RStudio Cache from unknowingly filling up a good chunk of the user's data directory.
What do you think @moravveji ?

source/software/r_package_management.rst
WouterVanAssche: Yes I will do that with a note above that it should be done from a compute node specifically.

source/leuven/services/openondemand.rst
```diff
@@ -419,14 +419,20 @@ For more general information, please refer to the `official JupyterLab documenta
 RStudio Server
 --------------
 
-This interactive app allows you to run an RStudio session as a compute job.
-You will be running RStudio with R version 4.2.1.
-For more information on how to use RStudio, check out the `RStudio official documentation`_.
+This interactive app allows you to run an RStudio session on the cluster. 
+In the 'Toolchain year and R version' drop-down menu, you can control the R installation
```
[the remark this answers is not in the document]
WouterVanAssche: I agree that for module <names/version> it is nice to have it with back-quotes as it makes things easier to spot in a text and copy.

source/software/r_package_management.rst
WouterVanAssche: Agreed. But solved this missmatch slightly differently:
Since `$VSC_HOME` has limited quote, it is not..."

source/software/r_package_management.rst
WouterVanAssche: I implemented this suggestion, with the note that it should be done from a wice compute node.

source/leuven/services/openondemand.rst
moravveji: This is a good idea. If we manage to change the RStudio default behavior (not saving a closed session), then we can stick to forwarding the caches to VSC_DATA using `XDG_DATA_HOME` consistently across this page.
But, the question is whether we can change the RStudio default behavior.

I looked up `/apps/leuven/rocky8/icelake/2022b/software/RStudio-Server/2022.07.2+576-foss-2022b-Java-11-R-4.2.2/R/ServerOptions.R`, and there I find:

```R
# set save defaults for high-performance (only if the administrator
# or user hasn't set them explicitly already)

if (is.null(getOption("save.defaults")))
   options(save.defaults=list(ascii=FALSE, compress=FALSE))

if (is.null(getOption("save.image.defaults")))
   options(save.image.defaults=list(ascii=FALSE, safe=TRUE, compress=FALSE))
```
But, this is too vague to me; the comment line is also ambiguous and I wonder if this is what we are looking after. Furthermore, the passed values are mainly `FALSE` which do not give extra room to turn off auto-saving (if that is what these two lines mean).

So, any other idea to turn off auto saving of sessions?

moravveji: In RStudio, it is 'easy' to override the defaults and suppress loading of previous sessions upon startup. It is just under 'Tools -> Global Options -> General -> RSessions'.

However, I looked up the source code for the equivalent of these, so that we can perhaps modify these defaults in `Options.R` and/or `ServerOptions.R`. But, so far, I have not spotted from where they actually come from. Can you perhaps ask this from the Posit support?

source/leuven/services/openondemand.rst
```diff
@@ -436,16 +442,20 @@ You can do this by using the ``lib`` argument for both the ``install.packages``
...
-  To avoid this, you can redirect this cache to your data directory by setting ``$XDG_DATA_HOME``
-  variables in your ``~/.bashrc``.
+  To avoid this, you can redirect this cache to your data directory by setting the ``$XDG_DATA_HOME``
+  variable in your ``~/.bashrc``:
 
   .. code-block:: bash
 
     echo "export XDG_DATA_HOME=$VSC_DATA/.local/share" >> ~/.bashrc
 
+- Additionally, it is advised to change the default behaviour of RStudio to not restore .RData
```
moravveji, line 455: `.RData` in two places

source/leuven/services/openondemand.rst
```diff
@@ -436,16 +442,20 @@ You can do this by using the ``lib`` argument for both the ``install.packages``
...
+  into the workspace on start up and to never Save the workspace to .RData on exit.
+  You can do this via the RStudio interface:
```
moravveji, line 457: ... via the RStudio interface by going to '...'. 
please add the terminating point to the sentence

moravveji (approved):
Besides the last two cosmetic changes I requested, this PR is good to go.
Thanks WouterVA for the whole lot.

Issue #427, opened 2024-08-20
Lirias virtual collection label


LucIJspeert:
In trying to add my publication to the Virtual Collection of VSC in Lirias, I found that neither "VSC", nor “High Performance Computing” are valid labels. The latter is suggested on the FAQ site. Instead, I presume the label I found, "KUL-HPC", is the correct one to use.

I would suggest to add this in the documentation page referencing Lirias, here:
https://docs.vscentrum.be/how_do_i_acknowledge_the_vsc_in_publications.html

stevenvdb:
Thanks for letting us know, the update has been done in https://github.com/hpcleuven/VscDocumentation/pull/428.

PR #428, opened 2024-08-23
Update Lirias virtual collection to KUL-HPC

stevenvdb:
As noted in https://github.com/hpcleuven/VscDocumentation/issues/427, the name of the virtual collection in Lirias is called "KUL-HPC" and not “High Performance Computing" (see https://research.kuleuven.be/nl/associatienet/output/lirias/faqs_webnl/overzicht-virtuele-collecties-overview-virtual-collections). It looks like this has changed during the migration to Lirias 2.0, see https://admin.kuleuven.be/icts/english/services/wms/admin/staff/migration-to-lirias-2.0.

### DIFF
--- a/source/how_do_i_acknowledge_the_vsc_in_publications.rst
+++ b/source/how_do_i_acknowledge_the_vsc_in_publications.rst
@@ -21,6 +21,6 @@ English version
   Foundation - Flanders (FWO) and the Flemish Government.
 
 |KUL| Moreover, if you are in the KU Leuven association, you are also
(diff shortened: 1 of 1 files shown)


moravveji (approved):
Thanks for the catch.

jooghe (approved):
Good report of the user. We need to check the webmasters of the Lirias page as wel because the referenve https://www.kuleuven.be/onderzoek/associatienet/ondersteuning/output/lirias/faqs_webnl/overzicht-virtuele-collecties-overview-virtual-collections does not exist ... .

PR #432, opened 2024-10-07
Cloud: Auto-apply variables docs

GoodOldJack12:
Updated docs for https://github.com/hpcugent/openstack-templates/pull/21

Requires user communication

### DIFF
--- a/source/cloud/manage_volumes.md
+++ b/source/cloud/manage_volumes.md
@@ -6,7 +6,11 @@ volume to a running instance or detach a volume and attach it to another
 instance at any time. You can also create a snapshot from a volume, or
 delete it.
 
-## Create a volume
+## Terraform
+The terraform templates support volume provisioning for VMs.
+Since DATE, it can also mount and resize them automatically. See {ref}`volumes`
+
+## Create a volume manually
 
 1.  Open the Volumes tab and select the Volumes category.
 
--- a/source/cloud/terraform.md
+++ b/source/cloud/terraform.md
@@ -97,7 +97,7 @@ file in a public place.
 chmod 600 ~/.config/openstack/clouds.yaml
 ```
 
-## Terraform modules
+## Basic VM configuration
 :::{tip}
 If you are **not** using the VSC login node, you need to make sure to:
 1) [Install Terraform](https://developer.hashicorp.com/terraform/install)
@@ -114,8 +114,7 @@ Navigate to the environment directory first:
 cd ~/openstack-templates/terraform/environment
 ```
 
-### Basic VM configuration
-This module deploys one VM with a public IP address and can be configured with extra options
+The `single.tf.example` deploys one VM with a public IP address and can be configured with extra options
 Copy the `single.tf.example` code to the `main.tf` file like so:
 ```shell
 cat examples/single.tf.example >> main.tf
@@ -204,9 +203,9 @@ You can add additional access rules with the `access_rules` variable:
     }
   ]
 ```
-:::{warning}
+```{warning}
 Setting custom rules disables the default rule.
-:::
+```
 :::
 
 ## Deploy Terraform templates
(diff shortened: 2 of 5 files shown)


source/cloud/terraform.md
```diff
@@ -1,5 +1,7 @@
 # Orchestration Using Terraform
-
+:::{warning}
```
alvarosimon, line 2: I think we can just add this in the email and not in the docs

source/cloud/manage_volumes.md
```diff
@@ -6,7 +6,11 @@ volume to a running instance or detach a volume and attach it to another
 instance at any time. You can also create a snapshot from a volume, or
 delete it.
 
-## Create a volume
+## Terraform
+The terraform templates support volume provisioning for VMs.
+Since DATE, it can also mount and resize them automatically. See {ref}`volumes`
```
alvarosimon, line 11: I would remove since... and just say what it does now. We can add this into an email to the current users

source/cloud/terraform.md
```diff
@@ -134,24 +136,20 @@ The file now contains the definition for one VM with several options you can cus
...
-| public_image | The OS image for the public instances (and private instances by default)|
-| cluster_name | The basename for the vms. Public instance will be called `MyCluster-public` and private instances `MyCluster-private-x` |
 
+More advanced options are described further on. 
+### Private VMs
+You can also deploy a public VM and multiple private VMs (without a public IP) by defining multiple instance modules and setting `public = false` on the private VMs.
+You can then reach those VMs through your public VMs with ssh-agent forwarding (`ssh -A`). See `frontend_backend_nfs.example` for a complete example.
+
+:::{warning}
+Automated features like auto-mounting volumes will **not** work on private VMs and are automatically disabled.
```
alvarosimon, line 151: This could be confusing for users, at least at this point in the doc, maybe it should be just a note or included later it

source/cloud/terraform.md
```diff
@@ -347,6 +346,38 @@ EOT
 :::{warning}
 Exposing ports to the internet is potentially dangerous. Make sure that the application you're exporting is properly secured.
 :::
+(terraform_share)=
+## NFS Share
+The latest templates include a module to easily configure an NFS share.
```
alvarosimon, line 351: I would include an example in the templates about this and add here  ``cat xxx >> main.tf`` if they want to use it

source/cloud/terraform.md
```diff
@@ -265,68 +263,69 @@ plan (only directly from OpenStack).
...
-If you enabled these options upon initial creation of the VM, you don't have to do these steps
+## Automated variables
+The terraform templates include the ability to automatically adapt your VM based on certain variables.
+This is accomplished by installing scripts on the VM when it is first created.
+It requires ssh access to the VM.
+
+These features can be disabled by setting `scripts_enabled=false`.
+
+:::{warning}
+Automated variables will not work if you started using the templates before 11/2024.
```
alvarosimon, line 274: I would remove this from the docs as the previous warnings about the template changes. I would include this in the email for the current users not in the doc.

source/cloud/terraform.md
```diff
@@ -134,24 +136,20 @@ The file now contains the definition for one VM with several options you can cus
...
 
+More advanced options are described further on. 
+### Private VMs
+You can also deploy a public VM and multiple private VMs (without a public IP) by defining multiple instance modules and setting `public = false` on the private VMs.
+You can then reach those VMs through your public VMs with ssh-agent forwarding (`ssh -A`). See `frontend_backend_nfs.example` for a complete example.
+
+:::{warning}
+Automated features like auto-mounting volumes will **not** work on private VMs and are automatically disabled.
+:::
 ## Deploy Terraform templates
```
alvarosimon, line 153: also I would move this section (Deploy Terraform Templates)  at the end, if not users will try to run the commands right now and they should read the rest before doing that

alvarosimon (changes requested):
@GoodOldJack12 it looks great, I have just some minor comments. I have tried the new templates following the current docs without issues :+1:

wdpypere (approved):

PR #434, opened 2024-10-09
Adding memory-per-core info to hardware spec pages

moravveji:
In response to a request from user(s) in the user survey of UD2023, in this PR, the information about the default memory per core is added to the following hardware specification pages. 
- genius
- superdome
- wice

- [x] First, [this internal PR](https://gitea.icts.kuleuven.be/ceif-lnx/puppet-hpc-root/pulls/1604) has to be approved.

### DIFF
--- a/source/leuven/tier2_hardware/genius_hardware.rst
+++ b/source/leuven/tier2_hardware/genius_hardware.rst
@@ -22,7 +22,7 @@ Hardware details
   - 192 GiB RAM (:ref:`memory bandwidth and latency measurements <memory bandwidth and latency cascadelake tier2>`)
   - default memory per core is 5000 MiB
   - 200 GB SSD local disk
-  - partition ``batch/batch_long``,
+  - partition ``batch|batch_long``,
     specific Slurm :ref:`options <submit_genius_batch>` apply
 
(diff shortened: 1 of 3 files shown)


moravveji:
@MaximeVdB : I updated the three pages for genius, superdome and wice and 
- extended/corrected each block where needed. E.g. sometimes newly-defined partition names were missing
- the `gpu_a100_debug` partition is now split from `interactive` partition, because the two have different `DefMemPerCPU`, and in this way, the text is easier to read. Please give this another look.

If we change the `DefMemPerCPU` for the Genius `interactive` partition from 2000MiB to 5000MiB, I have to fix this accordingly.

MaximeVdB (changes requested):

moravveji:
> Go ahead, but maybe also check the quickstart page withe the examples of the submit options. To be consistent with the MB or MiB notation.

The accepted units for memory specifications by `sbatch|srun|salloc` commands is e.g. `k|m|g|t`. So, here we do not have the MB vs MiB shenanigans. I checked the quick start pages anyway, and the memory specifications in the examples looked fine to me.

MaximeVdB (approved):

PR #440, opened 2024-10-25
Malware-free download link for FileZilla

moravveji:
I got a nice headsup from a user who mentioned that the current FileZilla download link leads to a copy that contains ads and malware. As explained [in this link](https://www.reddit.com/r/HTML/comments/1d7ko9r/is_filezilla_a_safe_program/), the current download URL leads to the quick download which is a 'sponsored version' including adware and gets detected by security as malware. For users machines equipped with malware detection, this is quite a headache. 

Instead, I replaced the download link with the direct download page from FileZilla: https://filezilla-project.org/download.php?show_all=1

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -338,7 +338,7 @@ rst_prolog += """
 .. _Cygwin: https://www.cygwin.com/
 .. _Docker: https://www.docker.com/
 .. _docs.globus.org: https://docs.globus.org
-.. _download FileZilla: https://filezilla-project.org/download.php?type=client
+.. _download FileZilla: https://filezilla-project.org/download.php?show_all=1
 .. _Eclipse download page: http://www.eclipse.org/downloads
 .. _Eclipse packages download page: https://www.eclipse.org/downloads/packages/
 .. _Eclipse: https://www.eclipse.org/


stevenvdb (approved):
Looks good to me.

On a side note; users with a KU Leuven managed laptop can install FileZilla with the Software Center. This is site-specific, so I am not sure we need to include it in VscDocs. Maybe it is also worth checking that the Software Center provides the clean version.

lexming (commented):
I get the same tarball from both links with checksum:
```
577e3ec41df48372b5e103e7a4c5f210adc0758c78bc4581f7b636cd7a7d201d FileZilla_3.67.1_x86_64-linux-gnu.tar.xz
```

How is this new link safer?

stevenvdb:
> I get the same tarball from both links with checksum:
> 
> ```
> 577e3ec41df48372b5e103e7a4c5f210adc0758c78bc4581f7b636cd7a7d201d FileZilla_3.67.1_x86_64-linux-gnu.tar.xz
> ```
> 
> How is this new link safer?
> 
> _update_: To be clear, in one case I downloaded by clicking on the old link image
> 
> And the other by clicking on the new link image
> 
> I cannot tell if both are bad tarballs, or both are good.
> 
> Maybe this only affects non-Linux versions?

For Windows, the current link downloads `FileZilla_3.67.1_win64_sponsored2-setup.exe`, which is quarantined by Windows Security. With the new link, the downloaded file is `FileZilla_3.67.1_win64-setup.exe` and then I at least don't get complaints from Windows Security. So it indeed seems to only affect Windows downloads. The MacOS downloads are also identical.

PR #441, opened 2024-11-05
Minor OnDemand Jupyterlab update

MaximeVdB:
My editor also took the opportunity to get rid of trailing whitespaces on the OOD page, so just see a736a60d4a8f75c6094f9496a9dad4f2d607ef84 for the real content changes. The main motivation here was that searching for IPython was not yet giving (relevant) hits.


WouterVanAssche (approved):

source/leuven/services/openondemand.rst
```diff
@@ -273,14 +273,11 @@ The same applies for other choices of partitions on Genius or wICE clusters.
...
-
-   This section is prepared for the scheduled migration of JupyterLab on
-   Thursday 25 July 2024. Some features will be available only after the migration.
-
-With this app you can create or (re)run Jupyter Notebooks. This can be handy both for R and Python coding.
-One of the benefits of using JupyterLab is the flexibility it offers in customizing user environments,
-which we explain below.
+With this app you can write and run
+`Jupyter <official JupyterLab documentation>`_ notebooks containing
+annotated Python or R code (among other languages). IPython consoles are
```
moravveji, line 278: It would be useful to explicitly add 'Julia' to this list, too. At the moment, searching for Julia returns nothing, but in this way, a curious Julia user receives a hint that Jupyter also supports that.

moravveji (approved):
Besides one remark about Julia, this is good2go.

PR #444, opened 2024-12-03
KULeuven correct GPU resource proportions

moravveji:
Recently, a [`job_submit.lua` plugin](https://slurm.schedmd.com/job_submit_plugins.html) got deployed on our production sites to impose a new GPU usage policy. Through the plugin, we ensure that each job requests a maximum allowed proportion of cores and memory per every requested GPU. This PR adds a new section to the [specific Slurm instructions](https://docs.vscentrum.be/leuven/slurm_specifics.html) relevant for the KU Leuven site.

### DIFF
--- a/source/leuven/slurm_specifics.rst
+++ b/source/leuven/slurm_specifics.rst
@@ -218,45 +218,27 @@ per GPU is provided in the table below.
      - 16
      - 187200
 
-The following remarks apply when submitting jobs to the GPU partitions:
+The submit filter will issue a warning if a job requests more cores or memory per GPU
+than what is listed above. If this happens, please adjust the Slurm options accordingly
+for your future jobs.
 
-* If a job requests more cores or memory per every GPU as listed above, the job
-  will not be submitted to the queue.
-  Instead, an informative message will be sent to the standard error stream.
-* A job may request less cores and/or memory per GPU than the maximum limit in the table above.
-* Instead of specifying ``--mem`` or ``--mem-per-cpu``, a job may request ``mem-per-gpu``.
-  In that case, the maximum value specified for ``--mem`` applies to ``--mem-per-gpu``, too.
-* For restricting maximum memory for GPU jobs, one has to choose one of the ``--mem``, ``--mem-per-cpu``,
-  or ``--mem-per-gpu`` options.
-* For multi-GPU jobs, the multiple of resouces from the table above applies.
-  E.g. the maximum allowed resources for a two-GPU job on wICE ``gpu_a100`` partition would look like:
+As an example, suppose that you need two A100 GPUs for your calculation, with just
+one core per GPU but with as much CPU memory as you can get. Such a job can be
+submitted as follows:
 
   .. code-block:: bash
-
+     # This job will get less than 18 cores per GPU, so this requirement is satisfied
+     # It will receive 126000 MiB of CPU memory per GPU, which is the maximum
+     # we can get without getting the submit filter warning
      sbatch --account=lp_myproject --clusters=wice --partition=gpu_a100 \
-            --nodes=1 --ntasks=36 --gpus-per-node=2 --mem=252000m \
+            --nodes=1 --ntasks-per-node=2 --gpus-per-node=2 --mem=252000 \
             myjobscript.slurm
 
-  Similarly, multi-node multi-GPU jobs can take up the entire cores and memory of the nodes.
-  But, resources can be specified per node and device:
+For more examples of valid GPU jobs, have a look at the
+:ref:`Genius <genius_t2_leuven>` and :ref:`wICE <wice_t2_leuven>`
+quickstart guides.
 
-
-  .. code-block:: bash
-
-     sbatch --account=lp_myproject --clusters=wice --partition=gpu_a100 \
-            --nodes=2 --ntasks-per-gpu=18 --gpus-per-node=4 --mem-per-gpu=126000m \
-            myjobscript.slurm
-
-* Due to the Multi-Instance GPU (MIG) configuration of the Nvidia A100 GPUs on the
-  wICE ``interactive`` partition, specifying ``--gpus-per-node=1`` will result in
-  allocation of 1/7th of the physical device.
-  One cannot request any additional GPU instance from this partition.
-* Slurm supports `GPU sharding <https://slurm.schedmd.com/gres.html#Sharding>`_, and this
-  feature is enabled for all our GPUs.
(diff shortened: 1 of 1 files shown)


source/leuven/slurm_specifics.rst
```diff
@@ -173,3 +173,98 @@ A few notes on this feature:
...
+  One cannot request any additional GPU instance from this partition.
+* Slurm supports `GPU sharding <https://slurm.schedmd.com/gres.html#Sharding>`_, and this
+  feature is enabled for all our GPUs.
+  The maximum shards per each GPU is equivalent to the number of cores of the compute host.
+  When requesting GPU shards, *no* resource limits apply.
+  In this case, the user is supposed to request the same number of cores as the requested
+  GPU shards.
+  It is also adviced to leave out memory specifications, and rely on the default memory per core.
+* All the examples given in the :ref:`Genius <genius_t2_leuven>` and :ref:`wICE <wice_t2_leuven>`
+  quick start guides fully comply with the correct resource proportions.
```
MaximeVdB, line 270: - For now there's only a warning; non-compliant jobs are not (yet) rejected.
- From the above change it should already be evident that these are upper bounds (and so that jobs remain free to request less resources per GPU).
- We should keep in mind that users will only need to bother with these constraints if they've intentionally raised the CPU resources for their job (the default is one core per node and default memory per core, so that always complies). As a result, one example should be enough (at least for now).
- I don't see the added value of the interactive partition information in this context (and it's of course already covered in https://docs.vscentrum.be/leuven/wice_quick_start.html#submit-to-the-interactive-partition).
- Shards are a resource we only support in the context of trainings where it is needed to accommodate many users on a limited number of devices. Regular jobs/users are not supposed to use this feature and (IMO) should not even be made aware of it.

So based on this my suggestion is:

```suggestion
The submit filter will issue a warning if a job requests more cores or memory per GPU
than what is listed above. If this happens, please adjust the Slurm options accordingly
for your future jobs.

As an example, suppose that you need two A100 GPUs for your calculation, with just
one core per GPU but with as much CPU memory as you can get. Such a job can be
submitted as follows:

  .. code-block:: bash
     # This job will get less than 18 cores per GPU, so this requirement is satisfied
     # It will receive 126000 MiB of CPU memory per GPU, which is the maximum
     # we can get without getting the submit filter warning
     sbatch --account=lp_myproject --clusters=wice --partition=gpu_a100 \
            --nodes=1 --ntasks-per-node=2 --gpus-per-node=2 --mem=252000 \
            myjobscript.slurm

For more examples of valid GPU jobs, have a look at the
:ref:`Genius <genius_t2_leuven>` and :ref:`wICE <wice_t2_leuven>`
quickstart guides.

Aside from options such as ``--ntasks-per-node`` and ``--cpus-per-task``
(for CPU cores) and ``--mem`` and ``--mem-per-cpu`` (for CPU memory),
keep in mind that Slurm also offers options like ``--cpus-per-gpu`` and
``--mem-per-gpu``.
```

source/leuven/slurm_specifics.rst
```diff
@@ -173,3 +173,98 @@ A few notes on this feature:
   (MIG) is used. This is for instance the case on the wICE Slurm partition
   called ``interactive``. For jobs on that partition this feature is
   irrelevant.
+
+.. _gpu_cores_mem:
+
+GPU, memory and core proportions
+--------------------------------
```
MaximeVdB, line 180: I think this is a bit more descriptive:
```suggestion
.. _cpu_resource_limits_in_gpu_jobs:

CPU resource limits in GPU jobs
-------------------------------
```

MaximeVdB (changes requested):

avapirev (approved):

MaximeVdB (approved):

moravveji:
@MaximeVdB : after your approval, I still made few tiny changes (to fix build a build issue from applying suggestions), and I also paraphrased the last paragraph. So, I like to ask for your review, once again.

source/leuven/slurm_specifics.rst
```diff
@@ -173,3 +173,76 @@ A few notes on this feature:
...
+For more examples of valid GPU jobs, have a look at the
+:ref:`Genius <genius_t2_leuven>` and :ref:`wICE <wice_t2_leuven>`
+quickstart guides.
+
+Slurm offers advanced options for fine-grained resource specifications for GPU jobs.
+For instance, one may combine ``--ntasks`` and ``--cpus-per-gpu`` to limit the maximum
+cores per GPU.
+Similarly, one may specify the (minimum) CPU memory per GPU using the ``--mem-per-gpu``
+option.
+In either case, the maximum CPU limits from the table above shall be respected.
```
MaximeVdB: I don't see the added value of mentioning possibly confusing combinations like  ``--ntasks`` + ``--cpus-per-gpu``. The `In either case, ...` sentence is also unnecessary (in addition to sounding a bit archaic ;). So I'd prefer you to change this back:
```suggestion
Aside from options such as ``--ntasks-per-node`` and ``--cpus-per-task``
(for CPU cores) and ``--mem`` and ``--mem-per-cpu`` (for CPU memory),
keep in mind that Slurm also offers options like ``--cpus-per-gpu`` and
``--mem-per-gpu``.
```

MaximeVdB (changes requested):

MaximeVdB:
> Because the scope of this newly-added section is GPU-specific resources, I took out generic resource options such as `--mem|--mem-per-cpu|--cpus-per-task` etc. They belong to other pages.

This is about the ratio of CPU resources to GPU resources, so a juxtaposition of ``--cpus-per-gpu`` and
``--mem-per-gpu`` with ` `--mem|--mem-per-cpu|--cpus-per-task` is entirely appropriate ;).

source/leuven/slurm_specifics.rst
moravveji: improved in 04fffa0

moravveji:
> > Because the scope of this newly-added section is GPU-specific resources, I took out generic resource options such as `--mem|--mem-per-cpu|--cpus-per-task` etc. They belong to other pages.
> 
> This is about the ratio of CPU resources to GPU resources, so a juxtaposition of `--cpus-per-gpu` and `--mem-per-gpu` with ` --mem|--mem-per-cpu|--cpus-per-task` is entirely appropriate ;).

Agreed. I guess with 04fffa0, we reach a common ground. Is it so?

MaximeVdB (approved):
👍

PR #447, opened 2025-01-13
tier1 hortense: add debug_milan partition details

stdweird:


boegel:
@stdweird Can you clarify why this is closed?

PR #452, opened 2025-01-27
Reorganize documentation in services

lexming:
This PR reorganizes the documentation in services as discussed in the CUE work group:
* The 3 main categories stay the same: Compute, Data and Cloud
* Frontpage re-designed to show small cards for each service inside the 3 main categories
* Accounts and access converted to `VSC Accounts`. Now it only covers: Creation of accounts, Cryptographic Keys, MFA and Account Management
* Documentation on access methods through the terminal moved into new `Terminal Interface` section under Compute
* New service added to Compute for `Web Portals`
* Infrastructure split into two services `Tier-1 Infrastructure` and `Tier-2 Infrastructure`
* Added redirects for all old paths to their new location

The tooling to generate the documentation is updated as well. **Now tabs can be synced across tab panels.** This means that whenever the user selects a tab for their home institute, all other tab panels will automatically show that same tab. This does not only work for the opened document, but across the whole documentation site.

Another tooling improvement is the addition of a link checker. Now we can execute `make linkchk` and sphinx will build the documentation and check the status of all external links. We had many broken links, this PR fixes many many of them.

This new version of the documentation can be checked in https://vscdocumentationbeta.readthedocs.io/


source/accounts/generating_keys.rst
```diff
@@ -1,12 +1,11 @@
-#############
-Security Keys
-#############
+##################
+Cryptographic Keys
```
smoors, line 2: i would use the term `SSH Keys` instead, which i think is what most people use, and it's more clear those keys are for SSH authentication.

also, cryptographic keys don't necessarily mean ssh keys.

if you really want to use cryptographic, i would go for `SSH Cryptographic Keys`

lexming:
I synced this PR with current `master`.

@jooghe @moravveji @MaximeVdB: what is your feedback on these changes? can we get this moving in the short term?

source/accounts/lecturer_procedure_student_accounts_kuleuven_uhasselt.rst
```diff
@@ -1,8 +1,3 @@
-.. _lecturer procedure leuven:
-
-Lecturer's procedure to request student accounts (KU Leuven/UHasselt)
-=====================================================================
-
 We support using HPC for educational purposes, and we fully assist the lecturers
```
jooghe, line 1: it took me a while befor eI found the lnik in the navigation. But it's ok, I think.

jooghe (commented):
looks good to me, small changes in the content can be done after PR.

backelj:
lgtm

backelj (approved):
lgtm

PR #460, opened 2025-02-24
Update job_types.rst

toondehaene:
--parsable doesn't only print the jobnumber

see https://slurm.schedmd.com/sbatch.html#OPT_parsable

Fixed text and command

### DIFF
--- a/source/jobs/job_types.rst
+++ b/source/jobs/job_types.rst
@@ -250,12 +250,13 @@ For example, assume that we have two job scripts:
   It uses the environment variable ``perturbation_size`` to determine the perturbation to
   apply.
 
-To make ``sbatch`` print simply the job ID after submitting, use the ``--parsable`` option.
+To make ``sbatch`` print the job ID after submitting, use the ``--parsable`` option. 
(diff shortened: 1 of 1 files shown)


source/jobs/job_types.rst
```diff
@@ -250,12 +250,13 @@ For example, assume that we have two job scripts:
   It uses the environment variable ``perturbation_size`` to determine the perturbation to
   apply.
 
-To make ``sbatch`` print simply the job ID after submitting, use the ``--parsable`` option.
+To make ``sbatch`` print the job ID after submitting, use the ``--parsable`` option. 
+The --parsable option prints "jobnumber;cluster". We use cut to fetch just the jobnumber.
```
smoors, line 254: in brussel it just prints the job ID, so i would rephrase this a bit:
```suggestion

- In some clusters, the ``--parsable`` option also prints the cluster name in the form "<job ID>;<cluster>". We can use cut to fetch just the job ID.

```

lexming:
Due to the changes in https://github.com/hpcleuven/VscDocumentation/pull/452, this file has moved to `source/compute/jobs/job_types.rst`.

toondehaene:
Ok, closing and resubmitting changes at new path

PR #466, opened 2025-04-22
Update job_types.rst

toondehaene:
resubmission of PR #460 

--parsable doesn't only print the jobnumber on HPC Leuven

see https://slurm.schedmd.com/sbatch.html#OPT_parsable

Fixed text and command

### DIFF
--- a/source/compute/jobs/job_types.rst
+++ b/source/compute/jobs/job_types.rst
@@ -250,12 +250,15 @@ For example, assume that we have two job scripts:
   It uses the environment variable ``perturbation_size`` to determine the perturbation to
   apply.
 
-To make ``sbatch`` print simply the job ID after submitting, use the ``--parsable`` option.
+To make ``sbatch`` print the job ID after submitting, use the ``--parsable`` option. 
+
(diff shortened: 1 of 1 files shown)


source/compute/jobs/job_types.rst
```diff
@@ -250,12 +250,15 @@ For example, assume that we have two job scripts:
   It uses the environment variable ``perturbation_size`` to determine the perturbation to
   apply.
 
-To make ``sbatch`` print simply the job ID after submitting, use the ``--parsable`` option.
+To make ``sbatch`` print the job ID after submitting, use the ``--parsable`` option. 
+
+- In some clusters, the ``--parsable`` option also prints the cluster name in the form "<job ID>;<cluster>". We can use cut to fetch just the job ID.
```
lexming, line 255: Minor suggestion to make it more clear:
```suggestion
.. note::
   In some clusters the ``--parsable`` option also prints the cluster name with the job ID in the form ``<job ID>;<cluster>``. We can use the command ``cut`` to fetch just the job ID as shown below.
```

PR #467, opened 2025-04-23
Remove confirmation email step + add waiting time

WilleBell:
Afaik there is no confirmation email being sent and no further action is needed to create a new group


lexming (approved):
LGTM

PR #472, opened 2025-05-14
Document use of wrapper script for compiled MATLAB code

jan-de-laet:
When installing MATLAB using EasyBuild, it is advised not to add the paths to the MATLAB runtime libraries to LD_LIBRARY_PATH in the module file (see https://github.com/easybuilders/easybuild-easyblocks/pull/2981). But these runtime libraries are required when running comiled MATLAB code, and so LD_LIBRARY_PATH needs those paths when running compiled MATLAB code. The preferred solution is to use the wrapper script for a MATLAB executable that is generated by the MATLAB compiler when compiling MATLAB scripts (https://github.com/easybuilders/easybuild-easyblocks/issues/2965#issuecomment-1665317981). The wrapper expects a first argument that provides the rootdir of the MATLAB installation that is being used. With a MATLAB module, that rootdir is given by EasyBuild environment variable EBROOTMATLAB. 

Both UGent and VUB already recommend using the wrapper script in their site-specific MATLAB documentation (https://hpc.vub.be/docs/software/usecases/#batch-mode and https://docs.hpc.ugent.be/Windows/MATLAB/#matlab-job-script), but the discussion of running compiled MATLAB code in the general VSC documentation tacitly assumed that LD_LIBRARY_PATH was set correctly, such that the MATLAB executable could be invoked directly. This PR adds the use of the wrapper script and EBROOTMATLAB to that general documentation.

### DIFF
--- a/source/compute/software/matlab_getting_started.rst
+++ b/source/compute/software/matlab_getting_started.rst
@@ -218,7 +218,8 @@ In addition to the MATLAB executable (``main`` in this example), the compiler
 also generates a wrapper file (``run_main.sh`` in this example) that can be
 used to invoke the MATLAB executable. It sets environment variable LD_LIBRARY_PATH
 to make sure that the MATLAB runtime libraries can be found by the executable,
-and next runs the executable.
+and next runs the executable. (The compiler generates a few other files as well,
+these can be ignored.)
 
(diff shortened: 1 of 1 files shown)


smoors (approved):
lgtm

w-lampaert (approved):

boegel (approved):

PR #494, opened 2025-08-26
update mathworksservicehost protocol KUL

w-lampaert:


moravveji (changes requested):

moravveji (approved):
LGTM.
You may also consider the to take out the `-v` option from the `rm` command, because it just informs you about useless stuff. Instead the `-f` option returns exit code `0` even if the target folder does not exist. Read all these as a personal suggestion.

w-lampaert:
This has slipped my mind. @smoors , is this fine for you?

PR #496, opened 2025-08-27
Adding PyCharm

moravveji:
PyCharm is added to the KU Leuven OnDemand portal as of 1 September 2025. The proposed page provides relevant information for the users specific to using PyCharm on the KU Leuven Tier-2 clusters.

### DIFF
--- a/source/compute/portal/ondemand/pycharm.rst
+++ b/source/compute/portal/ondemand/pycharm.rst
@@ -51,7 +51,7 @@ For more advanced configuration options, see also the `JetBrains documentation
 
 .. note::
 
-   For long-running jobs or for jobs using more than one node (with e.g.
+   For long-running jobs or for jobs using more than one node (with e.g.,
    ``mpi4py`` or ``dask.distributed``), we recommend :ref:`submitting batch
    jobs <job submission>` instead.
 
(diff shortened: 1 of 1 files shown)


MaximeVdB (changes requested):
When just using PyCharm as an editor, there is indeed not much that needs mentioning.

But I'm pretty sure that quite a few users will want to use features like the Python console and environment creation/selection, and that still needs to be described in a better way:

1) As far as I can see, PyCharm does *not* ship with a Python installation. Instead, it seems to search for ones that are available in the container image (`/usr/bin/python3....`) and propose those by default. And just like we wouldn't recommend users to rely on the system's Python, I think it is important to steer users away from these.

2) So far I didn't yet manage to let PyCharm offer me a Python interpreter from a centrally installed Python module. Unless you manage to find a way, we would need to make clear that such setups won't work. Assuming that also the interpreter is run within the container, then this will probably pretty much never work.

3) Choosing one of my Conda environments (or create a new one) did seem to work though.

MaximeVdB (approved):
👍 

I suppose you will ask one more person to review?

moravveji:
> 👍
> 
> I suppose you will ask one more person to review?

Indeed. I wanted to ask for reviews one at a time.

PR #499, opened 2025-09-09
Update vscode-tunnel.rst to include KUL/UH sites

avapirev:


moravveji (changes requested):
I have few remarks concerning the KUL-specifics:

- The tunnel name that contains the user VSC id (the one limited to 20 chars) can be different between KUL and VUB; e.g. we use currently a template like `vsc-vscXXXXX-wice` which is different from that of VUB. This needs to be made clear in the documentation with site-specific tabs. 
  Well ... we still have to internally agree on the tunnel name, which will happen outside this repo
- The local VSCode terminal shows that the `ttyd` module belongs to the 2024a toolchain. This means that users who want to use the Python(-related) modules are restricted to the 2024a toolchain only. This must be made clear in a 'Warning' box
- There is a mention of using Microsoft account, but I tried with my github connection, and it did not bite. The copilot worked as expected. So, perhaps this knot can be loosened a bit
- We also need to clarify that at KUL, we do not offer GPUs for this specific app. Think of it as a 'Note' box

PR #503, opened 2025-09-22
update OOD docs for VUB

smoors:
- how to request more memory
- update modules loaded by default
- minor writing style changes

### DIFF
--- a/source/compute/portal/ondemand/jupyterlab.rst
+++ b/source/compute/portal/ondemand/jupyterlab.rst
@@ -50,8 +50,8 @@ listed Python module).
    .. tab-item:: VUB
       :sync: vub
 
-      A large number of widely used Python packages are available via the
-      following modules, which are loaded by default:
+      A large number of widely used Python packages are already available by
+      default. The following modules are always loaded with JupyterLab:
 
       - ``Python-bundle-PyPI`` - packages for general usage
       - ``Scipy-bundle`` - data science packages like ``scipy``, ``numpy``, ``pandas``
@@ -60,7 +60,8 @@ listed Python module).
 
 If you need additional Python packages, you can load extra modules with Python
 packages via ``module load`` commands in the 'Pre-run Scriptlet' of the
-resources form.
+resources form, or via the :ref:`software_modules_extension` in your JupyterLab
+session (if supported by the OnDemand portal).
 
 After launch, the ``Python 3 (ipykernel)`` kernel is automatically available in
 your JupyterLab session. Along with the standard Python library, it also gives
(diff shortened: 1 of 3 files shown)


source/compute/portal/ondemand/jupyterlab.rst
```diff
@@ -37,19 +37,35 @@ drop-down menu.  An example would be '2023a and
...
+
+      Furthermore, you may choose to tick one of the checkboxes to load popular
+      modules from the same toolchain, such as ``SciPy-bundle`` (for widely used
+      packages like ``scipy``, ``numpy``, ``pandas`` and more) and/or ``matplotlib``.
+
+   .. tab-item:: VUB
+      :sync: vub
+
+      A large number of widely used Python packages are available via the
+      following modules, which are loaded by default:
```
lexming, line 54: Minor suggestion for clarity:
```suggestion
      A large number of widely used Python packages are already available by default.
      The following modules are always loaded with JupyterLab:
```

source/compute/portal/ondemand/jupyterlab.rst
```diff
@@ -37,19 +37,35 @@ drop-down menu.  An example would be '2023a and
...
+
 
-Once you launch a JupyterLab session, a default kernel called ``Python 3
-(ipykernel)`` is already available in your session.  This kernel, in addition to
-the Python standard library, would enable using extra packages from
-``SciPy-bundle`` and/or ``matplotlib``, if you selected them in the resource
-form.
+If you need additional Python packages, you can load extra modules with Python
+packages via ``module load`` commands in the 'Pre-run Scriptlet' of the
+resources form.
```
lexming, line 63: Why not just use the `Lmod` panel in JupyterLab? that is much easier

source/compute/portal/ondemand/rstudio-server.rst
```diff
@@ -16,14 +16,19 @@ that will be loaded for your session (such as ``R/4.4.1-gfbf-2023b``).
...
+      If needed, the ``R-bundle-Bioconductor`` module can be loaded as well to
+      get access to many preinstalled bioinformatics packages.
+
    .. tab-item:: VUB
       :sync: vub
 
-      The ``R-bundle-CRAN`` module can be loaded on top of the base R module,
-      providing hundreds of frequently-used R packages.
+      A very large number of widely used R packages are available via the
+      following modules, which are loaded by default:
```
lexming, line 26: Same suggestion for clarity as with JupyterLab:
```suggestion
      A very large number of widely used R packages are already available by default.
      The following modules are always loaded with RStudio:
```

source/compute/portal/ondemand/jupyterlab.rst
smoors: leuven doesn't have the modules extension, and if you regularly load the same modules it's easier in the pre-run scriptlet, because OOD remembers that list for you.

but i'll add it as alternative 👍

PR #507, opened 2025-10-03
sharing folders via vsc groups

moravveji:


source/accounts/vsc_user_groups.rst
```diff
@@ -124,5 +125,43 @@ Working with file and directory permissions
...
+- Change the group name of the folders in the hierarchy one by one to ``lp_shared_data``:
+
+  ::
+
+     chgrp lp_shared_data $VSC_DATA
+     chgrp lp_shared_data $VSC_DATA/results
+     # change group name recursively below the top hierarchy you want to share
+     chgrp -R lp_shared_data $VSC_DATA/results/post-processed
+
+- Make sure that the shared directory has read and execute access to all folders underneat,
```
stevenvdb, line 153: This phrase is not very clear. I guess you mean that, in this example, `$VSC_DATA` and `$VSC_DATA/results` need to be readable and executable by the group `lp_shared_data`?

source/accounts/vsc_user_groups.rst
```diff
@@ -124,5 +125,43 @@ Working with file and directory permissions
...
+  Let's call this new group ``lp_shared_data``.
+- Invite your VSC collaborators to join the group ``lp_shared_data``.
+- Change the group name of the folders in the hierarchy one by one to ``lp_shared_data``:
+
+  ::
+
+     chgrp lp_shared_data $VSC_DATA
+     chgrp lp_shared_data $VSC_DATA/results
+     # change group name recursively below the top hierarchy you want to share
+     chgrp -R lp_shared_data $VSC_DATA/results/post-processed
```
stevenvdb, line 151: Referring to the setgid bit mentioned above can be useful here.

source/accounts/vsc_user_groups.rst
```diff
@@ -124,5 +125,43 @@ Working with file and directory permissions
...
+
+- Make sure that the shared directory has read and execute access to all folders underneat,
+  and at least read access to all files:
+
+  ::
+
+     chmod -R g+r $VSC_DATA/results/post-processed
+     find $VSC_DATA/results/post-processed -type d -exec chmod g+x {} +
```
stevenvdb, line 159: Maybe a stupid comment, but at some point it might be useful to mention that other users need to have the full path to the `$VSC_DATA` subdirectory that is being shared.

source/accounts/vsc_user_groups.rst
moravveji: of course. see fe27285

stevenvdb (commented):
I made some minor comments on your modifications. I think we need input from other sites to see if they agree with this example of sharing data; if not, we might need site-specific tabs.

source/accounts/vsc_user_groups.rst
```diff
@@ -124,5 +125,58 @@ Working with file and directory permissions
...
+
+- Depending on the internal agreement within the collaborating team, you may choose to additionally
+  allow the group members to add/remove files to/from the shared hierarchy. If so, the group
+  members are expected to be careful specifically in removing files. In that case:
+
+  ::
+
+     chmod -R g+w $VSC_DATA/collab/post-processed
+
+- If the collaborators are expected to overwrite the existing files and/or add new files/folders
```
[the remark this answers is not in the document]
moravveji: Sure. See 6909392.

source/accounts/vsc_user_groups.rst
```diff
@@ -124,5 +125,58 @@ Working with file and directory permissions
...
+         find /path/to/directory -type d -exec chmod g+s '{}' \\;
+
+
+.. _sharing_via_vsc_groups:
+
+Sharing files and folders with other VSC users
+----------------------------------------------
+
+VSC users can share files/folders with other VSC users via `VSC groups <vsc_groups>`.
+This applies to any file and folder stored in the `VSC storage locations <data location>`.
```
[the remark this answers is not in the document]
moravveji: In response to your comment, I had already added this sentence:

"However, we discourage using your home directory for such purposes"

Is that sufficient?

source/accounts/vsc_user_groups.rst
```diff
@@ -124,5 +125,63 @@ Working with file and directory permissions
...
-         [user@foo]# find /path/to/directory -type d -exec chmod g+s '{}' \\;
+         find /path/to/directory -type d -exec chmod g+s '{}' \\;
+
+
+.. _sharing_via_vsc_groups:
+
+Sharing files and folders with other VSC users
+----------------------------------------------
+
+VSC users can share files/folders with other VSC users via `VSC groups <vsc_groups>`.
```
lexming: I'm not sure this the correct approach:
* this can create an explosion of groups in the VSC account page
* the share depends on the permissions of the top level directory of `$VSC_DATA`, which for instance in our case it is controlled by us and we do reset those permissions on each sync operation with the VSC account page

I would say that Globus is the preferred way to share data between users that are not part of the same Project or VO.

@jooghe maybe we should discuss this in the CUE meeting

stevenvdb (commented):
Before completing my review, I think we need input from other sites (and even within our group) if advising users to share data with groups through $VSC_DATA is really what we want to do (also see Alex' comment).

moravveji:
> We discussed this in the last CUE meeting. This is our recommendation:
> 
> * Users that need to share data in a recurrent basis should preferably be part of the same Project/Virtual Organization. If that is not the case, they can use this solution based on groups.
> * Users that need a one-time data share should use Globus (probably the share link functionality is enough). The solution based on groups should not be used for one-time shares to avoid explosion of groups.

I would like to revive this PR after months.
@lexming : What do you think of using tabs to cover different practices for different sites? In that way, my proposed changes can still stay in this "generic" page, and yet point at different practices in each VSC site.
If you are not in favor of using tabs, then I close this PR and put it somewhere in our KUL-specific pages.
Thanks in advance.

PR #556, opened 2026-01-21
clarify standard Globus transfer as copy

kafetzakid:
Clarify that a standard Globus transfer is a copy, not a move.


mstfdkmn (commented):
Thanks. I left a comment. Also, I realized a broken link. Could you fix it?
image

PR #560, opened 2026-02-04
patch firewall KU Leuven

mselwa:
Preparation for new central firewall setup at KU Leuven  (not yet to be merged to master until almost applied)

### DIFF
--- a/source/accounts/authentication.rst
+++ b/source/accounts/authentication.rst
@@ -84,7 +84,7 @@ Location Access Restrictions
 
       Non-managed laptops
             There are several possibilities to connect to the KU Leuven VSC clusters:
-                  *  From all locations (connecting from VPN B zone, from other VSC universities, from other Belgian IP addresses and from abroad):open the firewall access (https://firewall.vscentrum.be) and request a certificate.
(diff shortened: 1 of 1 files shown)


jooghe:
> I suggest to **not** add a new section called `Additional Firewall layer` as that is ambiguous and a bit confusing. Instead we can rename the previous section from `Connections from Abroad` to something like `Location Access Restrictions`; and then explain what restrictions are in place with a tabbed table per site:
> 
> * KU Leuven clusters:
>   
>   * managed laptops
>   * unmanaged laptops
> * UAntwerp clusters:
>   
>   * restrictions from abroad based on IP (copy paste existing text in `Connections from Abroad`)
> * UGent clusters:
>   
>   * restrictions from abroad based on IP (copy paste existing text in `Connections from Abroad`)
> * VUB clusters:
>   
>   * restrictions from abroad based on IP (copy paste existing text in `Connections from Abroad`)

It is indeed not an additional firewall layer, only a local change in our KU Leuven firewall setup. It moves from the node to the network, but that's not relevant to the user. Changing the "Abroad" section to 'location access restriction' looks like a good solution.

jooghe:
Additionally also the MFA page needs to be changed.
The note about the connections from abroad needs to be changed in managed versus non-managed laptop.

PR #565, opened 2026-02-06
skip unnecessary intermediate step to the ondemand docs

smoors:


source/compute/portal/index.rst
```diff
@@ -1,10 +1,89 @@
...
+       * Tier-2 :ref:`Hydra <Hydra cluster>`
+
+.. _KU Leuven OnDemand: https://ondemand.hpc.kuleuven.be/
+.. _UGent OnDemand: https://login.hpc.ugent.be/
+.. _VUB OnDemand: https://portal.hpc.vub.be/
+
+You can log in using the credentials of your home institution or your VSC
+credentials. SSH keys are not required to use the web portal.
+
+General features
```
lexming: this one as well,

```suggestion
Portal features
```

lexming (commented):
Great change 😍 

It is not necessary to move all those files though. That will break a lot of links. We can keep all files in place and just:
1. modify `compute/portal/index.rst` as you already did
2. delete `compute/portal/ondemand/index.rst` and put a single redirect for it
3. prepend `ondemand/` to all TOC entries in `compute/portal/index.rst`

PR #569, opened 2026-02-18
Add section about how to use EESSI to the Leuven module system page

stevenvdb:
EESSI is now available on the KU Leuven/UHasselt Tier-2 cluster, so this PR adds some documentation on how to use it. I have a few remarks:

- Adding a section on EESSI will also be part of https://github.com/hpcleuven/VscDocumentation/pull/534, but since the completion of that PR will take some time, I decided to already add a section to the KU Leuven/UHasselt specific page about modules.
- The instructions are not entirely self-contained, they assume the user already has some knowledge of lmod for example. I think this is ok, since we currently still use locally installed modules as the default and we do not expect beginning users to make use of EESSI immediately.

@laraPPr @boegel While this PR only updates a KU Leuven page, your comments are of course very welcome :)

### DIFF
--- a/source/leuven/leuven_module_system.rst
+++ b/source/leuven/leuven_module_system.rst
@@ -257,13 +257,14 @@ up EESSI first make the locally installed modules unavailable:
    module --force purge
    # Make EESSI available as a module
    export MODULEPATH=/cvmfs/software.eessi.io/init/modules/
-   # Load the default EESSI module
-   module load EESSI
+   # Load an EESSI module
+   module load EESSI/2023.06
 
 These setup commands need to be executed in each session where you want to use
-EESSI. After this, you can use the ``module`` commands as described above, but
-now those commands will interact with EESSI instead of locally installed
-modules. For example:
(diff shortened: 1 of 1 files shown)


source/leuven/leuven_module_system.rst
```diff
@@ -217,3 +217,73 @@ Keep in mind that also ``/apps/leuven/common/modules/all`` is part of your
...
+
+Althought EESSI is installed on all nodes of the KU Leuven Tier-2 cluster,
+the EESSI modules are not available by default and some commands need to be
+run to set up EESSI in your session. Also note that modules from EESSI are not
+compatible with locally installed modules, to the recommended commands to set
+up EESSI first make the locally installed modules unavailable:
+
+.. code-block:: shell
+
+   # Force unload any cluster module, which is normally use to select the
```
w-lampaert: Perhaps having just the comment '# Force unload any cluster module' is enough? It could be a bit confusing since this command alone does not select the correct module path, and it is also not the place to fully explain that.

PR #570, opened 2026-02-23
KUL OOD Matlab version policy

moravveji:


source/compute/portal/ondemand/matlab.rst
```diff
@@ -58,6 +58,11 @@ Before you connect
 Launching MATLAB
 ~~~~~~~~~~~~~~~~
 
+|KUL|
+Only MATLAB releases after ``R2024b`` are made available.
```
MaximeVdB:
```suggestion
Currently MATLAB ``R2024b`` is the oldest available version for this app.
```

(better to use active voice; more correct phrasing; hinting that this may change in the future)

PR #572, opened 2026-02-25
Update planning for Genius decommissioning

MaximeVdB:


moravveji (approved):
These changes look good to me.

But, I wonder whether this PR is also offers a good opportunity to inform the users that the GPU nodes will also be migrated to Rocky9. For now, this announcement has never found its way to the VSCdocs, yet.

If you think that it does not fit into this page/PR, I definitely listen to your suggestion.

PR #573, opened 2026-02-26
Update the GIO instructions for Leuven drives

MaximeVdB:


jan-de-laet (approved):
Works nicely in the ood desktop, good riddance of NX!

PR #591, opened 2026-03-30
Update MFA page so it refers to the new location access restriction section

stevenvdb:
The "MFA login" page was not updated as part of https://github.com/hpcleuven/VscDocumentation/pull/560, this provides a small fix to rectify that.

### DIFF
--- a/source/accounts/authentication.rst
+++ b/source/accounts/authentication.rst
@@ -66,6 +66,8 @@ clusters.
 
    mfa_login   
 
+.. _auth_location_access_restrictions:
+
 Location Access Restrictions
 ============================
 
--- a/source/accounts/authentication.rst
+++ b/source/accounts/authentication.rst
@@ -77,16 +77,16 @@ Location Access Restrictions
    .. tab-item:: KU Leuven/UHasselt
       :sync: kuluh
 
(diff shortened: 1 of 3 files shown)


stevenvdb:
Ping @mselwa @jooghe

jooghe (commented):
It indeed does not make sense to keep this note here. 
But maybe we should put an additional note on the tier2_login_nodes. page referring to the location access restriction table. 
Now we are saying "Direct login using SSH is possible to all login infrastructure without restrictions."
better "Direct login using SSH is possible to all login infrastructure keeping in mind the location access restrictions" ?

stevenvdb:
> It indeed does not make sense to keep this note here. But maybe we should put an additional note on the tier2_login_nodes. page referring to the location access restriction table. Now we are saying "Direct login using SSH is possible to all login infrastructure without restrictions." better "Direct login using SSH is possible to all login infrastructure keeping in mind the location access restrictions" ?

Done in a507514f70b7ab9f00507c3806529cb3a112d250

stevenvdb:
Ping @jooghe @mselwa

jooghe (approved):
Fine for me now.

Issue #550, opened 2026-01-15
[GSSI] Add/review sections on how Julia/Go/Perl/Rust on VSC


stevenvdb:
Add or review the following pages from the `restructure_and_update_software_section` branch:
- `source/compute/software/installing_software/perl_package_management.rst`
- `source/compute/software/installing_software/julia.rst`
- `source/compute/software/installing_software/perl_package_management.rst`
- `source/compute/software/installing_software/go.rst`

boegel:
During GSSI training, only Julia & R were covered, so let's focus on those.

For Perl, we can revisit what's now at https://docs.vscentrum.be/compute/software/perl_package_management.html (by @stevenvdb), main thing there is to update module and mention `Perl-bundle-CPAN`

For Go & Rust, a dedicated page can be added by @WilleBell

stevenvdb:
Some notes for Go and Rust were added as part of https://github.com/hpcleuven/VscDocumentation/pull/578, which for me is sufficient to close this one.

PR #595, opened 2026-03-31
[GSSI] add documenation about managing Julia packages

lexming:
partial fix for #https://github.com/hpcleuven/VscDocumentation/issues/550

### DIFF
--- a/source/compute/software/installing_software/julia.rst
+++ b/source/compute/software/installing_software/julia.rst
@@ -12,7 +12,7 @@ How to find available Julia packages?
 * The software module called ``Julia`` provides a standard installation of
   Julia without any extra packages
 * Some specific Julia packages can be provided with their own software modules
-  (*e.g.* IJulia)
+  (*e.g.* ``IJulia``)
 
 Installing Julia packages
 -------------------------
--- a/source/compute/software/installing_software/julia.rst
+++ b/source/compute/software/installing_software/julia.rst
@@ -144,12 +144,20 @@ Julia environments can be created on top of a software module and be made to
 
       $ module load Circuitscape/5.12.3-Julia-1.9.2
 
-2. Create a new environment based on the module's environment
+2. Create a new environment based on the module's environment by copying the
+   files ``Manifest.toml`` and ``Project.toml`` from the module
 
    .. code-block:: console
 
-      $ base_project=$(julia -E 'Base.load_path()[end]')
-      $ cp -r "$(dirname ${base_project:1:-1})" myNewEnv
+      $ mkdir myNewEnv
+      $ base_project="$(julia -E 'Base.load_path()[1]')"
+      $ base_project_dir="$(dirname ${base_project:1:-1})"
+      $ cp "${base_project_dir}/Manifest.toml" myNewEnv/
+      $ cp "${base_project_dir}/Project.toml" myNewEnv/
+
+   For instance, for ``Circuitscape/5.12.3-Julia-1.9.2``
+   these environment files can be found in a directory such as
+   ``/apps/brussel/<OS>/<arch>/software/Circuitscape/5.12.3-Julia-1.9.2/environments/v1.9``.
 
 3. Check that the new environment does have the packages from the module
 
--- a/source/compute/software/installing_software/julia.rst
+++ b/source/compute/software/installing_software/julia.rst
@@ -46,17 +46,17 @@ created with the same major and minor version of Julia as the loaded module.
 For example, if you load ``Julia/1.10.3-linux-x86_64``, all project
 environments for version 1.10 will be usable.
 
-|Recommended| Software installations of Julia packages can be heavy. Installing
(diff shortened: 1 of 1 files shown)


source/compute/software/installing_software/julia.rst
```diff
@@ -3,4 +3,182 @@
...
+Julia is a dynamic, high-performance programming language that is particularly
+well-suited for scientific computing and numerical analysis.
+
+How to find available Julia packages?
+-------------------------------------
+
+* The software module called ``Julia`` provides a standard installation of
+  Julia without any extra packages
+* Some specific Julia packages can be provided with their own software modules
+  (*e.g.* IJulia)
```
stevenvdb: Maybe put IJulia between backquotes to have the same formatting as for the ``Julia`` module in the previous bullet point.

source/compute/software/installing_software/julia.rst
```diff
@@ -3,4 +3,182 @@
...
+      julia> using Pkg
+      julia> Pkg.add("PackageName")
+
+Your personal *depot* (located in ``~/.julia``) or any custom project
+environments in your account will continue to be usable as long as they were
+created with the same major and minor version of Julia as the loaded module.
+For example, if you load ``Julia/1.10.3-linux-x86_64``, all project
+environments for version 1.10 will be usable.
+
+|Recommended| Software installations of Julia packages can be heavy. Installing
```
stevenvdb: Suggestion to replace "be heavy" with "can take up a lot of disk space".

source/compute/software/installing_software/julia.rst
```diff
@@ -3,4 +3,182 @@
...
+many packages or a single package that pulls many dependencies can quickly fill
+the :ref:`storage quota of your home directory <quota>`.
+We recommend to move your personal *depot* to your ``$VSC_SCRATCH`` and link
+``~/.julia`` to it.
+
+.. code-block:: bash
+
+   mkdir -p ~/.julia
+   mv -i ~/.julia $VSC_SCRATCH/julia
+   ln -s $VSC_SCRATCH/julia ~/.julia
```
stevenvdb: Not sure if this is worth mentioning, but this can limit the usage across VSC sites, since the symlink points to the value of `$VSC_SCRATCH` at the moment the symlink is created (which probably will not work on other VSC sites). Maybe `$VSC_DATA` is a better option? I am not sure how much space Julia packages require, but I assume they fit in `$VSC_DATA` in most cases? Of course you can also argue that accessing software over a mount at another site is not a good practice...

source/compute/software/installing_software/julia.rst
```diff
@@ -3,4 +3,182 @@
...
+   Julia packages loaded through software modules like
+   ``Circuitscape/5.12.3-Julia-1.9.2`` will be usable in Julia, but they will
+   **not be part of your project environment**. This means that you are free to
+   install your own versions of any of the Julia packages provided by the
+   loaded software module.
+
+Julia environment on top of software module
+-------------------------------------------
+
+Julia environments can be created on top of a software module, inheriting all
```
stevenvdb: What does "inheriting" mean in this context? It looks like you need to copy part of the installed module into your project directory.

stevenvdb (changes requested):

source/compute/software/installing_software/julia.rst
lexming: fixed in https://github.com/hpcleuven/VscDocumentation/pull/595/commits/60ec2ca4a1618f5e1f8814209d5677b0f4af5c33

source/compute/software/installing_software/julia.rst
lexming: sure, made in https://github.com/hpcleuven/VscDocumentation/pull/595/commits/c10521de8e44aa645087306cbc467bc41214f689

source/compute/software/installing_software/julia.rst
lexming: that's a good point, I agree that `VSC_DATA` can avoid pitfalls when switching clusters and should not cause any significant difference for this installations.

Fixed in https://github.com/hpcleuven/VscDocumentation/pull/595/commits/c10521de8e44aa645087306cbc467bc41214f689

source/compute/software/installing_software/julia.rst
lexming: By "inherit" I mean that all Julia packages in the module will be present and usable in the Julia environment. The "copy" command is not copying the packages themselves just the definition files of the environment of the module. But I see that the commands I wrote can be confusing.

I have fixed those commands in https://github.com/hpcleuven/VscDocumentation/pull/595/commits/3a34a2a017ff8aec8e76cc93318c83b3952f2216. Now they unmistakably copy the definition files into the custom julia environment and nothing else.

stevenvdb (approved):
Looks good to me!

RVerschoren (approved):
Looks good to me as well!

PR #597, opened 2026-03-31
[GSSI] Refresh page on Perl package management

stevenvdb:
- Use recent Perl module
- Recommend to also load Perl-bundle-CPAN
- Set PERL5LIB instead of relying on local::lib
- Use VSC environment variables when configuring cpan

Partially solves https://github.com/hpcleuven/VscDocumentation/issues/550

### DIFF
--- a/source/compute/software/installing_software/perl_package_management.rst
+++ b/source/compute/software/installing_software/perl_package_management.rst
@@ -3,8 +3,6 @@
 Perl package management
 =======================
 
-TODO-GSSI
-
 Introduction
 ------------
 
@@ -18,13 +16,11 @@ ensure that your code can be run on any platform that supports Perl.
 
 However, many useful extensions to and libraries for Perl come in the
 form of packages that can be installed separately. Some of those are part
-of the default installation on VSC infrastructure.
+of the Perl modules available on VSC infrastructure.
 
 Given the astounding number of packages, it is not sustainable to
-install each and everyone system wide. Since it is very easy for a user
-to install them just for himself, or for his research group, that is not
-a problem though. Do not hesitate to contact support whenever you
-encounter trouble doing so.
+install each and everyone system wide. :ref:`This section <perl-install-own-packages>`
+explains shows how users can install Perl packages themselves.
 
 Checking for installed Perl packages
 ------------------------------------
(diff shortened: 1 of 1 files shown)


source/compute/software/installing_software/perl_package_management.rst
```diff
@@ -36,12 +32,19 @@ those in your ``PERL5LIB`` environment variable.
 
 #. Load the module for the Perl version you wish to use, e.g., ::
 
-      $ module load Perl/5.28.2-foss-2018a
+      $ module load Perl/5.40.0-GCCcore-14.2.0
 
-#. Run cpan: ::
+#. In general it is recommended to also load the corresponding
```
boegel, line 37:
```suggestion
#. In general it is recommended to also (or only) load the corresponding
```

source/compute/software/installing_software/perl_package_management.rst
```diff
@@ -50,9 +53,9 @@ this purpose, the ``cpan`` utility first needs to be configured. Replace
 the paths ``/user/leuven/301/vsc30140`` and ``/data/leuven/301/vsc30140``
```
boegel: No way to do this via environment variables?

source/compute/software/installing_software/perl_package_management.rst
stevenvdb: I clarified this by adding
```
(note it is also sufficient to only load this module, since the correct
   ``Perl`` module will be loaded automatically as a dependency)
```

source/compute/software/installing_software/perl_package_management.rst
stevenvdb: It turns out this is possible; the old comment in the cpan instructions "Note that you can not use an environment variable for the path." no longer holds. The instructions have been updated to make use of `${VSC_HOME}` and `${VSC_DATA}`.

source/compute/software/installing_software/perl_package_management.rst
```diff
@@ -36,45 +32,55 @@ those in your ``PERL5LIB`` environment variable.
...
 
-      cpan[2]> o conf makepl_arg INSTALL_BASE=/data/leuven/301/vsc30140/perl5
-
-   Note that you can not use an environment variable for the path.
+      cpan[2]> o conf makepl_arg INSTALL_BASE=${VSC_DATA}/perl5
 
 #. Set the preference directory path: ::
 
-      cpan[3]> o conf prefs_dir /user/leuven/301/vsc30140/.cpan/prefs
+      cpan[3]> o conf prefs_dir ${VSC_USER}/.cpan/prefs
```
MaximeVdB:
```suggestion
      cpan[3]> o conf prefs_dir ${VSC_HOME}/.cpan/prefs
```

MaximeVdB (changes requested):
Looks good! Just one comment

source/compute/software/installing_software/perl_package_management.rst
stevenvdb: Of course, not sure how I missed that.

MaximeVdB (approved):

PR #601, opened 2026-04-03
minor conda updates

smoors:
- `source activate` is deprecated: change to `conda activate`, see [blog post](https://www.anaconda.com/blog/how-to-get-ready-for-the-release-of-conda-4-4#:~:text=As%20the%20year%20winds%20down,conda%20deactivate%E2%80%9D%20your%20conda%20environments)

- add example job script using `conda run`
- minor writing style changes

@MaximeVdB feel free to modify this PR as you see fit.

### DIFF
--- a/source/compute/software/installing_software/conda_based_managers.rst
+++ b/source/compute/software/installing_software/conda_based_managers.rst
@@ -38,10 +38,10 @@ installed ``Miniforge3`` module, for example:
 
 This has the following advantages:
 
-* The same base installation does not get duplicated by every user.
-* You will not be able to add packages to the ``base`` environment,
-  which is generally considered good practice.
-* By default, Miniforge will only include the (unrestricted) ``conda-forge``
+* The base installation is shared, so it isn’t duplicated for each user.
+* You cannot install additional packages into the ``base`` environment,
+  enforcing the recommended practice of keeping it unchanged.
+* By default, Miniforge only includes the (unrestricted) ``conda-forge``
   community channel and not the default channels which are subject to
   :ref:`Anaconda's terms of service <conda_channels>`.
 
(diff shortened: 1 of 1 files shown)


MaximeVdB:
> source activate is deprecated: change to conda activate, see [blog post](https://www.anaconda.com/blog/how-to-get-ready-for-the-release-of-conda-4-4#:~:text=As%20the%20year%20winds%20down,conda%20deactivate%E2%80%9D%20your%20conda%20environments)

Have you tried this yourself? Both with a local Miniconda installation and with a Miniforge3 module I have always needed to run `source activate ...` because otherwise `CondaError: Run 'conda init' before 'conda activate'` ensues. And as far as I'm aware the `conda init ...` part is something we don't like users to do as it essentially enables Conda by default and results in that conda initialize block in `~/.bashrc`. Personally I don't have strong preferences in either direction, but if you want to move to `conda activate ...` I think we'll need to ask if the others are fine with the implications.

smoors:
> > source activate is deprecated: change to conda activate, see [blog post](https://www.anaconda.com/blog/how-to-get-ready-for-the-release-of-conda-4-4#:~:text=As%20the%20year%20winds%20down,conda%20deactivate%E2%80%9D%20your%20conda%20environments)
> 
> Have you tried this yourself? Both with a local Miniconda installation and with a Miniforge3 module I have always needed to run `source activate ...` because otherwise `CondaError: Run 'conda init' before 'conda activate'` ensues. And as far as I'm aware the `conda init ...` part is something we don't like users to do as it essentially enables Conda by default and results in that conda initialize block in `~/.bashrc`. Personally I don't have strong preferences in either direction, but if you want to move to `conda activate ...` I think we'll need to ask if the others are fine with the implications.

that's why you first have to source the conda.sh script:
```bash
source $(conda info --base)/etc/profile.d/conda.sh
```

i'm ok if you prefer to keep `source activate`, it's just a suggestion.

i fully agree users should not do `conda init`, would be good to add a note about this.

MaximeVdB:
> ```shell
> source $(conda info --base)/etc/profile.d/conda.sh
> ```
> 
> i'm ok if you prefer to keep `source activate`, it's just a suggestion.
> 
> i fully agree users should not do `conda init`, would be good to add a note about this.

OK I glanced over that additional source command, perhaps because it was not in the place I expected it to be. I think `source $(conda info --base)/etc/profile.d/conda.sh` before every environment activation would be too cumbersome, but it seems we can include it in the suggested `loadminiforge` bash alias, so I will push some changes to that effect. Plus a note about `conda init`.

MaximeVdB (approved):
@smoors As far as I'm concerned you can merge if you also approve of the commits I added

smoors (commented):
looks good to me!

PR #611, opened 2026-05-06
Retire NX

stevenvdb:
Since NX is no longer available, I propose to remove the corresponding documentation.


source/accounts/mfa_login.rst
```diff
@@ -127,7 +127,6 @@ SSH Client name                        Purpose              Operating System
 :ref:`OpenSSH <OpenSSH access>`        text-based terminal  Linux, macOS
 :ref:`PuTTY <terminal putty>`          text-based terminal  Windows
 :ref:`MobaXterm <terminal mobaxterm>`  text-based terminal  Windows
-:ref:`NoMachine <NX start guide>`      graphical desktop    Windows, Linux, macOS
```
MaximeVdB, line 130: `source/redirects.list` also has a line referring to `nx_start_guide`. That doesn't seem to affect anything, but we might as well remove it.

MaximeVdB (changes requested):
Thanks for taking care of this - just one minor comment below

source/accounts/mfa_login.rst
stevenvdb: Done in https://github.com/hpcleuven/VscDocumentation/pull/611/commits/5c1ae2e1cf5ae153e69bfdfa5635c5914e8cb88c

MaximeVdB (approved):

PR #616, opened 2026-05-13
Update note about turbo frequencies on Rocky 8

stevenvdb:
This PR updates the note about the difference in turbo frequencies on Rocky 9 and Rocky 8, so it is clear there no longer are nodes on Rocky 8.


stevenvdb:
Ping @moravveji

moravveji (approved):

PR #620, opened 2026-05-18
Update Mindwell status (open pilot phase)

MaximeVdB:
(only to be merged once everyone has been added to the pilot account)


moravveji (approved):
LGTM.
Before starting the open pilot, we have to also inform users about the state of Mindwell via a separate PR.

PR #625, opened 2026-06-03
Mindwll in (leuven) docs

moravveji:


source/compute/portal/ondemand/interactive-apps.rst
```diff
@@ -137,8 +137,9 @@ your job resources:
    .. tab-item:: KU Leuven/UHasselt
       :sync: kuluh
 
-      Documentation on resources is available for both :ref:`Genius
-      <running_jobs_on_genius>` and :ref:`wICE <running jobs on wice>`.
+      Documentation on resources is available for :ref:`Genius
+      <running_jobs_on_genius>`, :ref:`wICE <running jobs on wice>` and
+      :ref:`Mindwell <running jobs on mindwell>`.
 
       In most cases we recommend using the ``interactive`` partition on wICE for
```
stevenvdb: Does this need an update as well? Or do we keep recommending wICE as the first choice for OnDemand?

source/compute/portal/ondemand/interactive-apps.rst
stevenvdb: For wICE -> For wICE and Mindwell

stevenvdb: This will also need an update, because on the Mindwell interactive partition, the GPU type is different.

stevenvdb (changes requested):
Largely ok of course. There are a few more things that might need to be changed in source/compute/portal/ondemand/interactive-apps.rst.

source/compute/portal/ondemand/interactive-apps.rst
moravveji: Hmmm ... This is subjective I guess, but relevant, and I think we have not explicitly discussed this among our team.
My personal reasoning is by mentioning our newest machine, this page (and other similar ones) will be valid for longer time.
But, we can always revise such after having a consensus from our team.

stevenvdb: Just to be clear, my comment is about the sentence:

>In most cases we recommend using the ``interactive`` partition on wICE for the interactive apps.

I don't have a strong opinion, but I would go for:

>In most cases we recommend using the ``interactive`` partition on wICE or Mindwell for the interactive apps.

, unless we have a reason to prefer users to keep using wICE.

source/compute/portal/ondemand/interactive-apps.rst
moravveji: does da404eff do the job?

source/compute/portal/ondemand/interactive-apps.rst
moravveji: Sure. At the cost of additional lines, I tried to make a high-level distinction between the interactive GPUs on wICE and Mindwell in b16c1025. Does it do the job?

source/compute/portal/ondemand/interactive-apps.rst
```diff
@@ -164,13 +164,17 @@ your job resources:
...
+      partition (e.g. a NVIDIA H100 for ``gpu_h100`` on wICE).
+      
+      For wICE and Mindwell, you can also request a GPU from the ``interactive`` partition.
+      The wICE interactive GPUs are virtually partitioned as 7 MIGs (Multi-Instance GPU).
+      In this configuration, each GPU slice provides roughly 1/7th of CUDA cores and memory
+      of an A100 GPU.  The interactive partition only allows you to request max 1 GPU (slice)
+      though.
+
+      Each Mindwell interactive node offers two NVIDIA RTX 5000 (Ada), which can be best
+      taken as whole for visualization purposes and lightweight machine learning applications.
```
stevenvdb, line 177: We have sharding enabled for these GPUs. Is it perhaps good to use only a shard of a GPU for interactive apps? Of course, we would need to support this in the OnDemand form first, so not blocking the current PR.

stevenvdb (approved):

source/compute/portal/ondemand/interactive-apps.rst
moravveji: I also came across similar thoughts when I was addressing the Ada devices.

We certainly have to discuss this among ourselves first, but here are my two cents:

- If I recall correctly, the purpose of `shard`ing configuration was to support educational resources, when needed
- So far, the VSCdocs are completely quiet about it, and I did not want to bring it up in this tiny PR. It deserves its dedicated PR, and sufficient discussion around it
- To my understanding, `shard`ed resource is not withing cgroups meaning that a "clumsy" user can oversubscribe the device and crash other users running application on the same device assuming that multiple users concurrently request shards. That's why I am stepping away from promoting it, for now
- (To be tested) I have an impression that when a single shard is occupied on a device, the rest of the device cannot be fully allocated for another job/user, leading to possible waste of resources (i.e. the remaining/unallocated shards)

So, all in all, let's first talk about it among ourselves to shortlist specific usecases before promoting it and adapting e.g. OOD app forms for it.

PR #632, opened 2026-06-26
Disclaimer about HPC for education

moravveji:
- KUL admin may refuse to reserve resources
- There is a risk of unavailability in case of hardware issues

### DIFF
--- a/source/_static/css/disclaimer-block.css
+++ b/source/_static/css/disclaimer-block.css
@@ -1,8 +1,8 @@
 .admonition.disclaimer {
-    border-left: 4px solid red;
+    border-left: 4px solid #f25224;
 }
 
 .admonition.disclaimer .admonition-title {
-    background-color: red;
+    background-color: #f25224;
     color: white;
 }
--- a/source/_static/css/disclaimer-block.css
+++ b/source/_static/css/disclaimer-block.css
@@ -1,0 +1,8 @@
+.admonition.disclaimer {
+    border-left: 4px solid red;
+}
+
+.admonition.disclaimer .admonition-title {
+    background-color: red;
+    color: white;
+}
--- a/source/accounts/hpc_for_education.rst
+++ b/source/accounts/hpc_for_education.rst
@@ -23,7 +23,7 @@
       .. admonition:: Disclaimer
          :class: disclaimer
 
-         Requests for reserving resources will be processed by the local support team.
+         Requests for reserving resources will be processed by the HPC support team.
          It may happen that for technical or administrative reasons, the request cannot
          be granted.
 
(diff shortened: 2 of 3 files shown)


source/accounts/hpc_for_education.rst
```diff
@@ -20,6 +20,17 @@
...
+      .. admonition:: Disclaimer
+         :class: disclaimer
+
+         Requests for reserving resources will be processed by the local support team.
+         It may happen that for technical or administrative reasons, the request cannot
+         be granted.
+
+         More importantly, the VSC cannot guarantee that the reserved resources be
+         available during the reservation in case of unforeseen circumstances,
+         such as hardware failures or urgent maintenance.
```
mselwa: Maybe also add
critical security vulnerability

mselwa (approved):
one more type of uregtn issues is critical security vulnerability - maybe also mention?

source/accounts/hpc_for_education.rst
moravveji: good idea. done in b4c39884.

source/accounts/hpc_for_education.rst
```diff
@@ -20,6 +20,17 @@
       - reserving compute resources (only if neccessary and upon decision by site admins) 
       - technical support during the trainings (either on-site or remote)
 
+      .. admonition:: Disclaimer
+         :class: disclaimer
+
+         Requests for reserving resources will be processed by the local support team.
```
MaximeVdB:
```suggestion
         Requests for reserving resources will be processed by the HPC support team.
```

(I think some people may be confused by the word 'local', thinking of their own department IT or so)

source/_static/css/disclaimer-block.css
```diff
@@ -0,0 +1,8 @@
+.admonition.disclaimer {
+    border-left: 4px solid red;
+}
+
+.admonition.disclaimer .admonition-title {
+    background-color: red;
```
MaximeVdB: This standard 'red' does not go well with the rest. Could you try something else, perhaps something yellow or orange from the [VSC palette](https://www.vscentrum.be/presskit)?

MaximeVdB (changes requested):

source/accounts/hpc_for_education.rst
moravveji: sure. done in a77ea9f7.

source/_static/css/disclaimer-block.css
moravveji: why no! I picked the middle orange from the secondary template. See a77ea9f7 plz.

MaximeVdB (approved):

PR #634, opened 2026-07-01
add KULUH info for matlab setup

w-lampaert:
Made a new PR for the now-closed https://github.com/hpcleuven/VscDocumentation/pull/494, since it was outdated and quite messy due to rebase I did afterwards. @smoors I aligned the VUB text with ours (so that's all you need to review), but I can revert the changes if you prefer.

### DIFF
--- a/source/compute/portal/ondemand/matlab.rst
+++ b/source/compute/portal/ondemand/matlab.rst
@@ -31,6 +31,30 @@ Who can use MATLAB
 Before you connect
 ~~~~~~~~~~~~~~~~~~
 
+MATLAB automatically generates several hidden folders in your ``$VSC_HOME``,
+including the folder ``.MathWorks``, which can become quite big over time. To
+prevent ``$VSC_HOME`` from filling up, we recommend creating symlinks to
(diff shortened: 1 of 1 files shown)


smoors:
@w-lampaert i'm fine with the VUB changes.

the paragraph about refraining from running MATLAB on the login nodes applies to VUB as well. maybe we can move all the common things out of the tabs and only keep the para on `R2024b` in the KULUH tab?

w-lampaert:
I have considered the same thing, but not sure if those things would apply for Antwerp and Gent as well (but I would expect it though)

smoors:
@backelj @RVerschoren what do you think?

for Gent it's not an issue because they have their own documentation

backelj:
The changes to "Before you connect" are also fine for us, so I guess it's the same for all sites.

W.r.t. "Who can use MATLAB": currently we have it under the `antwerpenall` group (much like `brusselall`), but we are considering a change here (either create a group `al_matlab` or `al_uantwerp[en_something]` to limit the access even more).

smoors (approved):
lgtm

moravveji (approved):
LGTM

PR #650, opened 2026-08-18
add comment about multiple commands in pre-run scriplet

w-lampaert:


source/compute/portal/ondemand/interactive-apps.rst
```diff
@@ -49,7 +49,9 @@ job will be queued.
    * - Pre-run scriptlet
      - Add optional shell commands to your job before launching the app.
        For example, loading extra modules that you need within the app, sourcing a
-       specific script, or defining specific environment variables.
+       specific script, or defining specific environment variables. For multiple commands,
```
smoors: the problem with multiple lines in a `text_area` widget (which we use for the pre-run scriptlet) is that CR characters are added to each line.

at VUB we solve this by removing the CR characters, see here for example: https://github.com/vub-hpc/ondemand-vub/blob/1d4e8f411c693d8cb216cbb880370ca3f5588969/apps/paraview/template/script.sh.erb#L32

w-lampaert: This seems like a much better solution. Thanks for the suggestion!

